PR #4, opened 2022-12-05
Trajectory fixes

tompe17:
Trajectory action now works, tested on a Spot. I had to change to a MultiThreaded executor to get it to work properly.


tompe17:
Oh, I did not know all my commits was going to be visible. Let me know if that is a problem...

PR #10, opened 2023-01-06
Changes from BDAI

jbarry-bdai:
This PR contains all of the changes BDAI has made to the spot_ros2 driver.

**Features Added**
1. Multiple robots: A large number of these changes were driven by us needing to be able to run multiple robots together.  When the robot name is specified, the driver is run in the namespace of the robot's name and all frames are prefixed with the robot name.

2. Automatically converted protobuf messages: We wrote a tool to automatically convert protobufs to ROS messages.  This gives us a 1-1 ROS API.  We added an action corresponding to the `RobotCommand` proto, which lets us easily convert many of the Boston Dynamics examples to run through ROS.

3. Debian package for `spot_msgs`.

4. World object support: We added world objects to TF and a service to query for their types.

5. Taking the lease: We take the lease instead of acquiring it so that you don't have to release control from the tablet etc first.

6. "Mock" mode for spot_ros2: We added a simple mock mode to spot_ros2 that allows it to run without an actual Spot.

**Changes For Discussion**
1. We found that having the spot driver run an estop client resulted in us sometimes not knowing how to estop the robot.  We removed all calls to `resetEStop` from the driver and added a check at startup that there's an external estop client.  If this is not desirable behavior, we can make it optional.
**Resolution:** Option for taking estop

3. We added support on `bosdyn_msgs` and made `spot_msgs` a debian package.  To get these you must run:
```
wget -q -O /tmp/ros-humble-bosdyn-msgs_0.0.0-0jammy_amd64.deb https://github.com/bdaiinstitute/bosdyn_msgs/releases/download/v0.0.0-humble/ros-humble-bosdyn-msgs_0.0.0-0jammy_amd64.deb \
  && sudo dpkg -i /tmp/ros-humble-bosdyn-msgs_0.0.0-0jammy_amd64.deb \
  && rm /tmp/ros-humble-bosdyn-msgs_0.0.0-0jammy_amd64.deb

wget -q -O /tmp/ros-humble-spot-msgs_0.0.0-0jammy_amd64.deb https://github.com/bdaiinstitute/spot_msgs/releases/download/spot_msgs-v0.0.0-0/ros-humble-spot-msgs_0.0.0-0jammy_amd64.deb \
  && sudo dpkg -i /tmp/ros-humble-spot-msgs_0.0.0-0jammy_amd64.deb \
 && rm /tmp/ros-humble-spot-msgs_0.0.0-0jammy_amd64.deb 
```
If this is unacceptable, we could create a ROS debian (so it can be apt got) or we could move things relying on bosdyn_msgs out of the spot_ros2 driver (but I think a lot of our future development will rely on these messages so that may impact our collaboration :( ).
**Resolution:** `spot_msgs` debian package added to the repository and `README` updated with install instructions.

3. We added a simple `SpotCommander` that just does the service calls because we branched before the existence of `spot_command_driver`.  We use it both on the commandline and import it into our scripts.  This could be moved out of the `spot_ros2` repo or renamed if it's too confusing.
**Resolution:** BDAI-specific `SpotCommander` was moved out of `spot_ros2` repo.

4. Because we have multiple spots, using the `spot_login.yaml` file didn't work for us.  Instead we use environment variables (like BD does) to specify everything.  If this doesn't work for you all, we can add an option to read login information.
**Resolution:** Option to use either environment variables or parameters.  YAML files moved out of the repo and specified in the launch file, but an example file left in.  `README` updated with instructions.

5. We increased the rates in the driver slightly to get smoother action when we're following a fiducial.
**Resolution:** Increased rates

**Notes for Reviews**:
* spot_ros.py changes are not shown by default - those should be reviews
* conversions.py is not shown - that is an automatically generated file to convert the ROS and bosdyn messages back and forth.  it does not need to be reviewed.  manually written conversions live in "manual_conversions.py" which is imported with `*` into conversions.py

**Todo with pushing**
- [ ] Update BDAI documentation on running 

Tests:
- [x] No robot name
  - [x] Launching with no robot name
  - [x] rviz/tf
  - [x] tf_static isn't publishing all the time
- [x] BDAI tests
  - [x] Running nogui estop through ROS
  - [x] Running gui estop through ROS
  - [x] Running spot and estop
  - [x] Running formation follow
  - [x] rviz/tf with multiple robots
  - [x] spot commander


tompe17:
I found it very convenient to not have to run an estop client.  I just stopped the driver with ctrl-c to make the Spot stop.

So if it can be an option to have the old behaviour it would be nice.  But maybe that is not working now with the change that the tablet can be connected at the same time as the driver is running.

jbarry-bdai:
> I found it very convenient to not have to run an estop client. I just stopped the driver with ctrl-c to make the Spot stop.
> 
> So if it can be an option to have the old behaviour it would be nice. But maybe that is not working now with the change that the tablet can be connected at the same time as the driver is running.

Sure it can.  How would you all feel about moving `spot_ros2.yaml` out of the repository and making its file path an argument to the launch file to make it easy to set these up two different ways?

goch:
2: As long as the bosdyn_msgs  are in a separate repository and we are able to install them without the need to setup the spot_ros2 driver on machines that only want to read the messages, this should be fine.

4: Is it necessary to have both SpotCommander and spot_command_driver? Having 2 nodes that have similar functionality and similar names is indeed confusing. In the long run it's probably useful to merge the functionality of both nodes into one or rename them so that it is obvious where the difference is.

5: seems smart to set login information with environment variables As long as the variables are documented in the setup instructions. Storing them in yaml files just increases the risk of pushing it into git by accident or makes handling of multiple robots unnecessary complex

METEORITENMAX:
2: Same as goch. But when i see it right, then the bosdyn_msgs pkg is built with the spot_msgs package. We need a standalone repo msgs package that is cloned along the spot_ros2 driver. The msgs repo can then be separately cloned or install.

So i think the action and srv files and the left msgs of the spot_msgs repo should be integrated in bosdyn_msgs. And then we could remove the spot_msgs package and only use the bosdyn_msgs. 

And I think it's even more comfortable to install the messages this way instead of building them from sources yourself.

jbarry-bdai:
Thinking about the message setup, I made spot_msgs and bosdyn_msgs two separate packages and repositories.  My reasoning here is that I'd like to have a place to define new custom spot msgs without having to download all of bosdyn_msgs.  It's also easy to make them one mega-repo though if folks strongly prefer that!  @METEORITENMAX @jiuguangw @goch

spot_driver/spot_driver/spot_wrapper.py
```diff
@@ -464,7 +520,7 @@ def releaseEStop(self):
 
     def getLease(self):
         """Get a lease for the robot and keep the lease alive automatically."""
-        self._lease = self._lease_client.acquire()
+        self._lease = self._lease_client.take()
```
jbarry-bdai, line 523: Note this change.  This will take the lease even if something else has it, which made switching between ROS and tablet operation smoother for us.  If this is a problem, it can be a variable!

spot_driver/spot_driver/spot_wrapper.py
```diff
@@ -501,32 +557,48 @@ def _robot_command(self, command_proto, end_time_secs=None, timesync_endpoint=No
             id = self._robot_command_client.robot_command(lease=None, command=command_proto, end_time_secs=end_time_secs, timesync_endpoint=timesync_endpoint)
             return True, "Success", id
         except Exception as e:
+            self._logger.error("Unable to execute robot command, exception was" + str(e))
             return False, str(e), None
 
     def stop(self):
         """Stop the robot's motion."""
+        response = self.claim()
```
jbarry-bdai, line 565: Note that we changed all of these to claim and/or power on as necessary.

METEORITENMAX (approved):

jiuguangw (approved):

jbarry-bdai:
@METEORITENMAX @jiuguangw i added some README stuff and reorganized a launch file a bit... I didn't realize that would require new approvals but it does...

METEORITENMAX (approved):

PR #17, opened 2023-02-02
refactor synchronized feedback to add arm and gripper commands

benaxelrod:


benaxelrod:
@jbarry-bdai please review

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -72,6 +72,11 @@
 MAX_DURATION = 1e6
 MOCK_HOSTNAME = "Mock_spot"
 
+COLOR_END    = '\33[0m'
+COLOR_RED    = '\33[31m'
+COLOR_GREEN  = '\33[32m'
+COLOR_YELLOW = '\33[33m'
```
jbarry-bdai, line 78: Instead of this use `self.node.get_logger().info/warn/error`.  They should show up colored - sometimes docker does dumb stuff with `tty` but we should fix that inside docker.  Anything that goes into the logger will also get put into the rosbag so that's important.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -421,18 +426,24 @@ def handle_max_vel(self, request, response):
 
     # -1 = failed, 0 = in progress, 1 = succeeded
     def _goal_complete(self, feedback):
+        FAILED = -1
+        IN_PROGRESS = False
+        SUCCESS = True
```
jbarry-bdai, line 431: Should probably be "SUCCEEDED for grammar but I don't really care

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -445,87 +456,128 @@ def _goal_complete(self, feedback):
...
+            # IN_PROGRESS.  And if it makes it to the bottom of the function, then all components
+            # must be satisfied, and it returns SUCCESS.
+            # One corner case to know about is that the commands that don't provide feedback, such
+            # as a velocity command will return SUCCESS.  This allows you to use them more effectively
+            # with other commands.  For example if you wanted to move the arm with some velocity
+            # while the mobility is going to some SE2 trajectory then that will work.
+
+            sync_feedback = feedback.command.synchronized_feedback
+            if sync_feedback.arm_command_feedback_is_set == True:
+                arm_feedback =  sync_feedback.arm_command_feedback
```
jbarry-bdai, line 499:
```suggestion
                arm_feedback = sync_feedback.arm_command_feedback
```

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -445,87 +456,128 @@ def _goal_complete(self, feedback):
...
+                if (arm_feedback.status.value != arm_feedback.status.STATUS_PROCESSING):
+                    return IN_PROGRESS
```
jbarry-bdai, line 501: this could probably use a comment because it's counter-intuitive

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -445,87 +456,128 @@ def _goal_complete(self, feedback):
...
+                elif (arm_feedback.feedback.feedback_choice == arm_feedback.feedback.FEEDBACK_ARM_JOINT_MOVE_FEEDBACK_SET):
+                    if (arm_feedback.feedback.arm_joint_move_feedback.status.value != arm_feedback.feedback.arm_joint_move_feedback.status.STATUS_COMPLETE):
+                        return IN_PROGRESS
+                elif (arm_feedback.feedback.feedback_choice == arm_feedback.feedback.FEEDBACK_NAMED_ARM_POSITION_FEEDBACK_SET):
+                    if (arm_feedback.feedback.named_arm_position_feedback.status.value != arm_feedback.feedback.named_arm_position_feedback.STATUS_COMPLETE):
+                        return IN_PROGRESS
+                elif (arm_feedback.feedback.feedback_choice == arm_feedback.feedback.FEEDBACK_ARM_VELOCITY_FEEDBACK_SET):
+                    # ArmVelocityCommand provides no feedback
+                    print(COLOR_YELLOW + 'WARNING: ArmVelocityCommand provides no feedback' + COLOR_END)
+                    pass
```
jbarry-bdai, line 514: probably worth noting that a pass will eventually result in a success return?

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -445,87 +456,128 @@ def _goal_complete(self, feedback):
...
+                elif (arm_feedback.feedback.feedback_choice == arm_feedback.feedback.FEEDBACK_ARM_GAZE_FEEDBACK_SET):
+                    if (arm_feedback.feedback.arm_gaze_feedback.status.value != arm_feedback.feedback.arm_gaze_feedback.STATUS_TRAJECTORY_COMPLETE):
+                        return IN_PROGRESS
+                elif (arm_feedback.feedback.feedback_choice == arm_feedback.feedback.FEEDBACK_ARM_STOP_FEEDBACK_SET):
+                    # Stop command provides no feedback
+                    print(COLOR_YELLOW + 'WARNING: Stop command provides no feedback' + COLOR_END)
+                    pass
```
jbarry-bdai, line 521: this should return SUCCESS i think

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -904,7 +956,7 @@ def step(self):
             self.node_rate.sleep()
 
 def main(args=None):
-    print('Hi from spot_driver.')
+    print(COLOR_GREEN + 'Hi from spot_driver.' + COLOR_END)
```
jbarry-bdai, line 959: This is ok to leave as a print because you don't have a node yet.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -1149,10 +1201,10 @@ def main(args=None):
             printed = False
             while spot_ros.spot_wrapper.is_estopped():
                 if not printed:
-                    print('\033[93mWaiting for estop to be released.  Make sure you have an active estop.'
+                    print(COLOR_YELLOW + 'Waiting for estop to be released.  Make sure you have an active estop.'
```
jbarry-bdai, line 1204: also ok to leave as print

jbarry-bdai (approved):
Thanks so much for doing this!  Minor comments.

spot_driver/spot_driver/spot_ros2.py
benaxelrod: I agree.  I think if this is anything other than `STATUS_PROCESSING` then we should probably return `FAIL` instead of `IN_PROGRESS`.  I think I was afraid that it might return `STATUS_UNKNOWN` for an iteration before `STATUS_PROCESSING`.  I'm a little new to the spot_driver, so I was also just following what the previous version did.

For reference, the other values in the enum are:
```
#/ Behavior execution is in an unknown / unexpected state.
int8 STATUS_UNKNOWN=0

#/ The robot is actively working on the command
int8 STATUS_PROCESSING=1

#/ The command was replaced by a new command
int8 STATUS_COMMAND_OVERRIDDEN=2

#/ The command expired
int8 STATUS_COMMAND_TIMED_OUT=3

#/ The robot is in an unsafe state, and will only respond to known safe commands.
int8 STATUS_ROBOT_FROZEN=4

#/ The request cannot be executed because the required hardware is missing.
#/ For example, an armless robot receiving a synchronized command with an arm_command
#/ request will return this value in the arm_command_feedback status.
int8 STATUS_INCOMPATIBLE_HARDWARE=5
```

@jbarry-bdai If you think it should actually return `FAIL`, let me know, it is an easy change.

spot_driver/spot_driver/spot_ros2.py
benaxelrod: we have a node at this point, so we could use the logger.  Do you still want a raw `print`?

spot_driver/spot_driver/spot_ros2.py
benaxelrod: What if you want to stop the arm while simultaneously moving mobility somewhere?  `pass`ing here means that you can still be running a mobility (and maybe gripper) command too.  and if not, then it will succeed at the bottom of the function.  

this is the logic i was trying to explain at the top of the function.  but probably could have been better.

spot_driver/spot_driver/spot_ros2.py
benaxelrod: Do you really want this line as the first line in the function, or can it be moved down a few lines to be under `create_node` so that we have the logger?

spot_driver/spot_driver/spot_ros2.py
jbarry-bdai: I had a reason for that but I forget what it was...  Let's leave it as a print for now.

spot_driver/spot_driver/spot_ros2.py
jbarry-bdai: I think this is ok to be left here so if node creation fails or something it's a little easier to track down.

spot_driver/spot_driver/spot_ros2.py
jbarry-bdai: Oops sorry I meant to delete this comment.

spot_driver/spot_driver/spot_ros2.py
jbarry-bdai: I think it returns `STATUS_UNKNOWN` right at the beginning before it goes into processing.

benaxelrod:
addressed a bunch of @jbarry-bdai 's comments.  untested.  will test next week.

spot_driver/spot_driver/ros_helpers.py
```diff
@@ -1,10 +1,12 @@
 import os
+import time
```
jbarry-bdai, line 2: This is my change and should be reverted - sorry!  Mistake while doing integration on Friday.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -421,18 +421,24 @@ def handle_max_vel(self, request, response):
...
         if not feedback:
-            return False
+            self.node.get_logger().error('ERROR: no feedback')
+            return IN_PROGRESS
+
         if feedback.command.command_choice == feedback.command.COMMAND_FULL_BODY_FEEDBACK_SET:
             if (feedback.command.full_body_feedback.status.value !=
                     feedback.command.full_body_feedback.status.STATUS_PROCESSING):
-                return False
+                return IN_PROGRESS
```
jbarry-bdai, line 435:
```suggestion
                return IN_PROGRESS  # This normally means the goal hasn't been acknowledged by the robot yet
```

jbarry-bdai (commented):
Looks good except for the thing I accidentally introduced :-p

benaxelrod:
tested.  @jbarry-bdai ready to merge

PR #23, opened 2023-02-15
Fix publishing of images

DavidWatkins:
The current publisher for RGB and depth images is slow. I have made the following improvements
- Added publishing the correct scoped frame_id for images
- Removed image publishing from spot_ros2.py and created a separate node to publish rgb, depth, or depth_registered
- Added arguments to publish depth_registered or depth aligned with the rgb frame

### DIFF
--- a/spot_driver/spot_driver/publish_cameras.py
+++ b/spot_driver/spot_driver/publish_cameras.py
@@ -74,12 +74,7 @@ def translate_ros_camera_name_to_bosdyn(camera_source: str, camera_type: str):
 
 class SpotImagePublisher(rclpy.node.Node):
     def __init__(self):
-        # super().__init__(f"{self._spot_name}_{self._camera_source}_{self._camera_type}_image_publisher")
         super().__init__(f"spot_image_publisher")
-        self.declare_parameter("camera_source", "")
-        self._camera_source = self.get_parameter("camera_source").value
-        self.declare_parameter("camera_type", "")
-        self._camera_type = self.get_parameter("camera_type").value
         self.declare_parameter('spot_name', '')
         self._spot_name = self.get_parameter('spot_name').value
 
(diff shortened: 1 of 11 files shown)


spot_driver/spot_driver/publish_cameras.py
```diff
@@ -0,0 +1,312 @@
...
+import bosdyn.client.util
+from bosdyn.api import image_pb2
+from bosdyn.client.image import ImageClient, build_image_request
+from google.protobuf.timestamp_pb2 import Timestamp
+from builtin_interfaces.msg import Time, Duration
+from spot_driver.ros_helpers import get_from_env_and_fall_back_to_param
+import rclpy.node
+
+
+ROTATION_ANGLE = {
```
jbarry-bdai, line 19: This is in degrees?  That's a little unusual :)

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -226,73 +223,6 @@ def WorldObjectsCB(self, results):
...
-            self.right_image_pub.publish(image_msg1)
-            self.right_image_info_pub.publish(camera_info_msg1)
-            image_msg2, camera_info_msg2 = getImageMsg(data[2], self.spot_wrapper)
-            self.left_depth_pub.publish(image_msg2)
-            self.left_depth_info_pub.publish(camera_info_msg2)
-            image_msg3, camera_info_msg3 = getImageMsg(data[3], self.spot_wrapper)
-            self.right_depth_pub.publish(image_msg3)
-            self.right_depth_info_pub.publish(camera_info_msg3)
-
-            self.populate_camera_static_transforms(data[0])
```
jbarry-bdai, line 274: how are these transforms getting published now?

jbarry-bdai (commented):
Before committing this I'd like to understand better if we can put it all in one place.  I don't love having so much duplicated code and authentication etc...

spot_driver/launch/point_cloud_xyz.launch.py
```diff
@@ -37,22 +37,21 @@
 
 import launch_ros.actions
 import launch_ros.descriptions
+from launch.actions import DeclareLaunchArgument
+from launch.substitutions import LaunchConfiguration, PathJoinSubstitution
```
jbarry-bdai, line 41: Have we tested this file?  (I haven't used it in its previous or current state.)

spot_driver/spot_driver/ros_helpers.py
```diff
@@ -127,20 +128,21 @@ def createDefaulCameraInfo():
...
         data: Image proto
-        spot_wrapper: A SpotWrapper object
     Returns:
         (tuple):
             * Image: message of the image captured
             * CameraInfo: message to define the state and config of the camera that took the image
     """
     image_msg = Image()
-    local_time = spot_wrapper.robotToLocalTime(data.shot.acquisition_time)
+    local_time = spot_wrapper.robot_to_local_time(data.shot.acquisition_time)
```
jbarry-bdai, line 143: While stylistically correct, this seems like maybe an unnecessary change for this PR?  I'm worried we'll miss a call to it somewhere.

jbarry-bdai (commented):
This looks good to me except for changing the robot to local function call - since that's a public function I think at this point we need to leave it the way it is because we don't know who might be using it outside of this.

I'd like to see someone from MASKOR review as well.

spot_driver/launch/point_cloud_xyz.launch.py
DavidWatkins: Yes it still works but it is very slow because the depth publishing is slow.

spot_driver/spot_driver/ros_helpers.py
DavidWatkins: I got all of them in the repository and the code still runs, but if you'd like me to undo it I will.

jbarry-bdai: I'm worried about external users like MASKOR.  @METEORITENMAX wdyt?

DavidWatkins: I went ahead an undid these changes.

jbarry-bdai (approved):

jbarry-bdai:
(Kept the branch because we're still pointing at it)

skpawar1305:
also, somehow CPU usage reduced when publishing images was done in a separate node.
I personally used to use it before https://github.com/skpawar1305/spot_ros2/blob/main/spot_driver/spot_driver/get_images.py
n it was always stable

PR #27, opened 2023-03-03
standlone arm urdf and rviz2 launch file and cofiguration

bilal-bdai:
adds a separate xacro file for the standalone arm with a base link for the purpose of loading in rigid body dynamics and adds the rviz2 launch file for visualization. Fyi, I kept the distance between the actual base origin of spot, and the base for this robot as is,


jbarry-bdai (approved):
Looks like a nice addition - thanks!

PR #28, opened 2023-03-25
Separate the spot wrapper from this repository, include as submodule

heuristicus:
This PR removes spot_wrapper and graph_nav_util from this repository to move it to a repository which is also used by spot_ros. The PR for the other repository is at https://github.com/heuristicus/spot_ros/pull/114.

You can find the wrapper repository at https://github.com/heuristicus/spot_wrapper, and the issue with outstanding tasks at https://github.com/heuristicus/spot_wrapper/issues/1. The only outstanding thing is to somehow merge the different image publishing methods. Regardless, everything should still function as before in terms of the driver.

You will need to install the spot_wrapper python module as per the readme:

```
git submodule init
git submodule update

pip3 install -e spot_wrapper
```

You should be able to check easily whether the module was correctly installed by running 

```
python3
from spot_wrapper.wrapper import SpotWrapper
```

To maintain behaviour of the driver, I've added two parameters. `use_take_lease` can be used to take the lease rather than acquiring it whenever a call to `claim` is made, and `get_lease_on_action` can be used to get the lease when performing an action which requires it, and also power on for stand, battery charge, and self-right.

I don't have a ros2 system to test with, so haven't been able to verify that this PR functions.

Tests for BDAI to run:

- [x] Ctrl C driver during while loops
- [x] Driver not holding estop doesn't error
- [x] Images are publishing at correct rate
- [x] Full system


jbarry-bdai:
Note spot wrapper review is going on at https://github.com/bdaiinstitute/spot_wrapper/pull/2

jbarry-bdai (approved):
@heuristicus please note that the git submodule here is updated to the `jlb/ros2_updates` branch in the `spot_wrapper` repo (https://github.com/bdaiinstitute/spot_wrapper/pull/3).

@METEORITENMAX please let us know if you would prefer we not merge this!  I've done the testing I want for BDAI.

jiuguangw (approved):

davidwatkins-bdai (approved):

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "spot_wrapper"]
+	path = spot_wrapper
+	url = git@github.com:heuristicus/spot_wrapper
```
skpawar1305, line 3: can you change url to https://github.com/bdaiinstitute/spot_wrapper
Thank you.

PR #31, opened 2023-04-01
Add examples and update documentation

jbarry-bdai:
This PR is to try to make the repo a little easier for folks to use.  Specifically:
* I added an install script and updated the README to instruct folks to use that
* I added an examples directory with examples of how to use the robot command action (because that's what we primarily do so those are the examples I have) and how to convert a BD example to use ROS2.
* I removed the requirement that spot driver launch have a config file (if you use the environment variables and are happy with the default parameters, it doesn't need one).
* I added the `ros_utilities` submodule, which has our wrappers for tf and action clients.  It also has the wrapper for action server, which is duplicated as well but I will fix that in a different PR.


examples/simple_arm_motion/README.md
```diff
@@ -0,0 +1,141 @@
...
+The commands are built the same way after this, but we change how we send them.  The Spot API code has a function for waiting until the arm has moved to position:
+```python
+        # Send the request
+        cmd_id = command_client.robot_command(command)
+        robot.logger.info('Moving arm to position 1.')
+
+        # Wait until the arm arrives at the goal.
+        block_until_arm_arrives_with_prints(robot, command_client, cmd_id)
+```
+In ROS2, we convert the created protobuf to a ROS2 action goal and we use the action client `send_goal_and_wait` function to replace the `block_until_arm_arrives_with_prints` function (we do not get the printing, but we could echo the action feedback topic or use the non-blocking `send_goal_async` if we wanted to do that):
```
amessing-bdai, line 127: "action client wrapper", so if someone skips down to here they aren't confused with the ROS2 ActionClient not having that function

examples/simple_arm_motion/README.md
```diff
@@ -0,0 +1,141 @@
...
+    # Send the request and wait until the arm arrives at the goal
+    node.get_logger().info('Moving arm to position 1.')
+    robot_command_client.send_goal_and_wait(action_goal)
+```
+
+The last thing to note in this example is the line
+```python
+    tf_listener.shutdown()
+```
+Owing to a bug in how tf threads are handled, if you want programs involving the TF listener wrapper to end cleanly, you have to make this call before the `tf_listener` is destructed.
```
amessing-bdai, line 141: Interesting. Not for this PR, but would it make sense to make a `__del__` function?

amessing-bdai (approved):

examples/simple_arm_motion/README.md
```diff
@@ -0,0 +1,141 @@
...
+* A spot commander: This is a [wrapper](../utilities/utilities/simple_spot_commander.py) around service clients that call the spot driver to do simple things like get the lease and stand.  This is used in place of calls like `blocking_stand`.
+* A robot command action client: This is the ROS2 action client that sends goals to the ROS2 action server (for more information about ROS2 actions see [here](https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Actions/Understanding-ROS2-Actions.html)).  This is used in place of the Spot API `RobotCommandClient`.  We use a [wrapper](https://github.com/bdaiinstitute/ros_utilities/blob/main/bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py) around the built in ROS2 action client that allows us to wait for the goal to return without risk of deadlock.
+
```
amessing-bdai, line 62: Not sure if it goes here, but maybe just add a small block about how to do this same thing if you are in a multi-spot config (basically just where to put the robot\_name/namespace

examples/simple_arm_motion/README.md
jbarry-bdai: It doesn't work!  I tried every way I could think of but it still hangs.  I suspect the problem is in some sort of order-of-delete thing...  You can mess with it locally with the `tf_listener_wrapper_test` and see if you can figure it out...

examples/simple_arm_motion/README.md
jbarry-bdai: I think I should probably write a multi-robot example instead...

amessing-bdai: Fair, that is the better option

amessing-bdai (approved):

PR #34, opened 2023-04-03
Declares hostname as a launch argument

ahoelscher-bdai:
This allows the hostname to be specified via an argument to the launch command, or via a child launch file that includes the spot driver with IncludeLaunchDescription. The `SPOT_IP` environment variable will override the argument if both are specified.


jbarry-bdai (commented):
I think MASKOR uses yaml config files to specify hostname (cuz they have only one robot!).  Can you test what happens with that?  Does the IP wind up coming from the default launch argument or from the config file?

PR #35, opened 2023-04-10
Add conversion functions for empty protobufs since they get called

jbarry-bdai:
Also fix adding to protobuf lists (should be `append` was `add`)


myeatman-bdai (approved):
Looks good to me.

amessing-bdai (approved):

PR #39, opened 2023-04-18
Add CI to spot_ros2 repo

pangtao-BDAI:
This PR adds the workflow that builds the docker image in which tests are run. The test workflow will be added in a separate PR.

PR #40, opened 2023-04-19
Add test workflow to CI

pangtao-BDAI:
This PR adds a workflow that builds this repo with a docker image built with #39. A few dependencies to the dockerfile are also added to make the build successful.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Test spot_ros2
+name: spot_ros2 CI
 on:
   pull_request:
   push:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: Test spot_ros2
```
jiuguangw, line 1: Please rename this "spot_ros2 CI". Normally it doesn't matter but in this case, I believe this string shows up in the Readme as the CI badge (so fairly obvious and should have a clear name).

pangtao-BDAI: Done.

jiuguangw (approved):
Thanks @pangtao-BDAI !

PR #45, opened 2023-05-01
[SW-106] Convert the image publisher node to use functions from ros helpers

amessing-bdai:
- [SW-102] Fix image_publishers CLI bug
- [SW-106] Convert the image publisher node to use functions from ros helpers
- [SW-104] Converts using the `|` operator for type hints to union as this allows earlier version of python (<3.10) to run the code

### DIFF
--- a/spot_driver/spot_driver/publish_cameras.py
+++ b/spot_driver/spot_driver/publish_cameras.py
@@ -18,6 +18,8 @@ from builtin_interfaces.msg import Time, Duration
 from spot_driver.ros_helpers import get_from_env_and_fall_back_to_param
 import rclpy.node
 
+from .ros_helpers import bosdyn_data_to_image_and_camera_info_msgs
+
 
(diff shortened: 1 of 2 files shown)


spot_driver/spot_driver/publish_cameras.py
```diff
@@ -18,6 +18,8 @@
 from spot_driver.ros_helpers import get_from_env_and_fall_back_to_param
 import rclpy.node
 
+from .ros_helpers import bosdyn_data_to_image_and_camera_info_msgs
```
davidwatkins-bdai, line 21: Do we have to use a relative import here?

davidwatkins-bdai (approved):
Looks a lot cleaner. Good job!

spot_driver/spot_driver/publish_cameras.py
amessing-bdai: Didn't work when I didn't use the relative import and we use the same one in `spot_ros2.py`

PR #46, opened 2023-05-01
Missing commit from PR #45

amessing-bdai:
This commit should have been in the previous PR and sets the spot_driver to use the same ros_helper function as the one that is now used by the image publisher


jbarry-bdai (commented):
Is the `bosdyn_data_to_image_and_camera_info_msgs` imported?  It doesn't look like it is...

jbarry-bdai:
> Is the `bosdyn_data_to_image_and_camera_info_msgs` imported? It doesn't look like it is...

Never mind missed the `*` import from `ros_helpers`

jbarry-bdai (approved):

PR #51, opened 2023-05-04
Fix publish_depth_registered overwritten in spot_ros2

kaiyu-zheng:
## Issue Description

When I add publish_depth_registered: True in spot_ros.yaml, this parameter on spot_ros2 is still False. 
```
$ ros2 param get /Gosu/spot_ros2 publish_depth_registered
Boolean value is: False
```
I tried adding a print statement inside spot_ros2, and indeed, I get False for `publish_depth_registered`.

However, for some other parameters, like `auto_claim`, auto_stand, I am able to change their values in spot_ros.yaml file and have them reflected in the spot_ros2 node.

 Similarly, I tried adding a print statement inside spot_ros2 to test for auto_stand and it works.

## Cause

In `spot_driver_with_namespace.launch.py` and `spot_driver.launch.py`, the default value of the parameter `publish_depth_registered` (False) overwrites the parameter set in the YAML configuration file.

## Solution

Remove `drivers_params` entirely, including the declared arguments. Enforce the configuration of `publish_x_args` through configuration file.  Exception is made for `spot_name` in  `spot_driver_with_namespace.launch.py`.

The general reason, I believe, is that in ROS, we should separate out parameters we want to be defined through the command line vs. defined through the configuration file.  In the case of spot_ros2,   `spot_name` should be declared on the command line (for quick change), and the rest go into the configuration file (potentially more stable).  This adds clarity to how to configure for spot_ros2 as well.

### DIFF
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -49,7 +49,7 @@ def generate_launch_description():
         executable='spot_ros2',
         name='spot_ros2',
         output='screen',
-        parameters=[config_file, driver_params]
+        parameters=[config_file]
     )
 
     params = {'robot_description': robot_desc}
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -39,11 +39,6 @@ def generate_launch_description():
         default_value="false",
     )
 
-    driver_params = {
-        'publish_rgb': publish_rgb,
-        'publish_depth': publish_depth,
-        'publish_depth_registered': publish_depth_registered
-    }
     spot_driver_node = launch_ros.actions.Node(
         package='spot_driver',
         executable='spot_ros2',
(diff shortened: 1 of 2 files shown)


amessing-bdai:
I think this is one option for fixing this. Another is to have the default for the `publish_rgb`, `publish_depth`, and `publish_depth_registered` variables be `None` or some kind of null representing value and then through substitutions remove them from the parameter list if when evaluated they are null. I think that is closer to the original intent of the launch file.

However, that idea assumes we want to keep the original intent. I agree with @kaiyu-zheng about making the `publish_X` args only in the config file. If you are choosing to publish these, then you likely have launched another node with a subscriber and this isn't as temporary as say a robot name.

@davidwatkins-bdai @jbarry-bdai Thoughts?

kaiyu-zheng:
> There may be a use case where I was to publish nothing in my experiment, but then quickly turn it on via command line to debug something. That likely isn't too different than modifying the config file, though.

I think in this case, you could also create a new config file for publishing nothing, and switching between the two config files. Because the "publish nothing" use case might be useful in general, and it is also stable. 

So a user could have multiple config files for a few different use cases or system setups.

davidwatkins-bdai (approved):
It's not ideal to be missing command line arguments, but to get the functionality we need for now it's good enough. Thanks for doing this.

PR #53, opened 2023-05-10
[SW-142] Fix logger in passed into SpotWrapper in spot_ros2

kaiyu-zheng:
Currently, 
```
spot_ros.logger = logging.getLogger('rosout')
```
is passed into SpotWrapper upon its creation.

This doesn’t work (logs in SpotWrapper aren’t being printed).

Proposing to use `rcutils_logger.RcutilsLogger(name=LOGGER_NAME)` and set LOGGER_NAME to be `spot_wrapper` for more clarity.


davidwatkins-bdai (approved):
Elegant fix to the problem

PR #56, opened 2023-05-15
[WUD-126] Myeatman/manipulation api

myeatman-bdai:
This adds a stub for adding support for the manipulation api. Tested on robot.

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -7,7 +7,7 @@ from std_srvs.srv import Trigger, SetBool
 from geometry_msgs.msg import Twist, Pose
 
 
-from bosdyn.api import geometry_pb2, trajectory_pb2, robot_command_pb2, world_object_pb2
+from bosdyn.api import geometry_pb2, trajectory_pb2, robot_command_pb2, world_object_pb2, manipulation_api_pb2
 from bosdyn.api.geometry_pb2 import Quaternion, SE2VelocityLimit
 from bosdyn.api.spot import robot_command_pb2 as spot_command_pb2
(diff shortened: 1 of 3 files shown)


jbarry-bdai (commented):
Looks like the right idea!

You can edit the robot command handling so that there's not so much copy paste code - I'm pretty sure we're currently the only people using that.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -1191,6 +1253,10 @@ def main(args=None):
                                                                spot_ros.handle_robot_command,
                                                                callback_group=spot_ros.group)
 
+        spot_ros.manipulation_server = ActionServer(node, Manipulation, 'manipulation',
+                                                              spot_ros.handle_manipulation_command,
+                                                              callback_group=spot_ros.group)
```
amessing-bdai, line 1258: Should this be conditional on the robot having an arm?

jbarry-bdai (commented):
This looks right but I think there's a lot of copy/pasted code from robot command.  Could we create one function that does the bulk of the work?

myeatman-bdai:
> This looks right but I think there's a lot of copy/pasted code from robot command. Could we create one function that does the bulk of the work?

I believe it's possible to create a function that shares some of the workload, but if you do the diff between the whole `handle_robot_command` function and `handle_manipulation_command` I believe there's less similarity than you're hoping for. If we had a 3 or more duplicated workflows, like an additional `handle_flying_command`, I'd be much more onboard. But I'd like to just move on to actually trying to grasp something in the diagnose-repair workflow.

jbarry-bdai (approved):

PR #62, opened 2023-05-24
fixing missing '.value' for mode_parent_odom_tf

skpawar1305:
now odom/vision correctly become parent of body


jbarry-bdai:
Oh this is definitely the right thing to do, but we need to test it because we've been working around having body be the parent of odom/vision for a while!

@myeatman-bdai do you have a chance to test this branch on our HIL test while you're working on other stuff?  if not i can find someone else!

@llee-bdai this probably won't break FtL but it would have made it simpler!

kaiyu-zheng:
Also, this may be of separate concern, can we not have the "frame_prefix" as part of the default value of that parameter (and other related parameters)?  It is inconvenient for a user to specify the frame_prefix, which should be managed by spot_ros2.

kaiyu-zheng:
> if you set that parameter to 'odom', it'll become patent of body we use 'vision', as it's much more accurate

Interesting. Isn't `odom` the default value already? (see the quoted lines)

skpawar1305:
> I'm attempting to test this.
> 
> Below is the TF frames with default value for `mode_parent_odom_tf`
> image
> 
> You can see that 'body' is the parent frame of 'odom'. 
> 
> But now.... I am confused. My understanding is changing `mode_parent_odom_tf` would somehow result in 'odom' being the parent frame of 'body'. But looking at the definition of that parameter...
> https://github.com/bdaiinstitute/spot_ros2/blob/ded66f5484564c98a49982f8e3a5ed589b6691dc/spot_driver/spot_driver/spot_ros2.py#L1077-L1078
> 
> it is set to somet string.  I am confused what value I should set it to to make `odom` be the parent of `body` -- What is `mode_parent_odom_tf` supposed to mean? 
> 
> Note that this isn't a question to you @skpawar1305 . But out of curiosity, what value did you set it to to achieve what you wanted (i.e. make `odom` be parent of `body`)?  That would help us testing.
> 
> 

if you set that parameter to 'odom', it'll become patent of body
we set it to 'vision', as it's much more accurate

skpawar1305:
> > if you set that parameter to 'odom', it'll become patent of body we use 'vision', as it's much more accurate
> 
> Interesting. Isn't `odom` the default value already? (see the quoted lines)

it's, but the issue is, without '.value', the variable is not set to 'odom', it's not a strong type :p

skpawar1305:
> Is this related to https://github.com/heuristicus/spot_ros/issues/95 at all? I'm not sure I fully understand the issue but it seems similar. 

no, that issue is regarding/odometry topic
we never use that,
this PR fixes tf2 tree

kaiyu-zheng:
So I am a bit confused.  Let me try to understand.

* Before this fix (i.e. before adding .value), body is the parent of odom by default.
* After this fix, odom is the parent of body by default.

But this fix has nothing to do with the default value of `mode_parent_odom_tf`, which is set to `odom`.  Something seems off.

skpawar1305:
> So I am a bit confused.  Let me try to understand.
> 
> * Before this fix (i.e. before adding .value), body is the parent of odom by default.
> * After this fix, odom is the parent of body by default.
> 
> But this fix has nothing to do with the default value of `mode_parent_odom_tf`, which is set to `odom`.  Something seems off.
>  
> 

default value is 'odom', but the variable isn't recognised as 'odom' without using .value

heuristicus:
Odom is the parent frame of body in the ROS version, I believe that is the default way the transform tree is constructed, since we are taking it directly from the tree returned by the robot state.

skpawar1305:
> > > So I am a bit confused.  Let me try to understand.
> > > 
> > > * Before this fix (i.e. before adding .value), body is the parent of odom by default.
> > > * After this fix, odom is the parent of body by default.
> > > 
> > > But this fix has nothing to do with the default value of `mode_parent_odom_tf`, which is set to `odom`.  Something seems off.
> > 
> > default value is 'odom', but the variable isn't recognised as 'odom' without using .value
> 
> But I thought that by default the parent frame for odom is body @jbarry-bdai 

if you want odom to be child of body, then you can set that param to vision.. then your other codes should work by default maybe?

jbarry-bdai:
OK so there's a few things happening here that I'll try to make clearer:
- When Spot publishes its tree, it has `body` as the parent of both `odom` and `vision`.  This is incorrect - the static frame should parent the moving frame.
- The spot driver code intends to fix this by reversing any transform that has `mode_parent_odom_tf` as the child transform.  It does this with the check `if inverse_target_frame == frame_name`
- However, we were passing in `inverse_target_frame = mode_parent_odom_tf` where `mode_parent_odom_tf` is a ROS Parameter class rather than a string.  Therefore the check always fails and the transform isn't reversed.  Hence `body` is the root of the TF tree.  This PR fixes that issue.

However, there's _still_  a problem here which is that `self.mode_parent_odom_tf` includes the frame prefix.  Therefore if we're running with the robot name as the frame prefix (which we always do), `self.mode_parent_odom_tf` is `<robot_name>/vision` or `<robot_name>/odom`.  So the check in `ros_helpers.py` (line 412 I think) needs to be updated to `if inverse_target_frame == spot_wrapper.frame_prefix + frame_name`.  This is why you didn't see the change in your testing @kaiyu-zheng.

Additionally, this same mistake is made just below this in `spot_ros2.py` (line 120).  Should we fix this at the same time?
```
            if self.mode_parent_odom_tf == self.spot_wrapper.frame_prefix + 'vision':
                odom_msg = GetOdomFromState(state, self.spot_wrapper, use_vision=True)
```

kaiyu-zheng:
That's clarifying. Thanks! Part of the confusion for me was what `mode_parent_odom_tf` means (the variable name).  

Just summarizing / suggesting what we should do in this PR:
* Update line 412 in ros_helpers.py to be `if inverse_target_frame == spot_wrapper.frame_prefix + frame_name`
* Remove `self.spot_wrapper.frame_prefix` in line 120 in spot_ros2.py
* Remove `frame_prefix` as part of the default value for `mode_parent_odom_tf`
* Incorporate the original fix in this PR that adds `.value` to `mode_parent_odom_tf`
* Either add in-line documentation for what `mode_parent_odom_tf` means, or improve the variable name. 

Thoughts?

jbarry-bdai:
This looks right now!  @kaiyu-zheng can you run your test again?  Thanks so much!

skpawar1305:
> 

we can merge this once tested, we can open  another PR if there are other important changes
as of now, it's identical to https://github.com/bdaiinstitute/spot_ros2/pull/70

kaiyu-zheng:
Sorry to keep nagging, but could you change the variable name to `preferred_odom_frame`? It's easier to understand and matches the pattern with `excluded_frames`. Thanks! @skpawar1305

PR #66, opened 2023-05-30
FeatureBranch: adding basic arm services

philipsgithub:
Hi, I just added ros services regarding the arm. the wrapper interfaces already exist in the spot_wrapper module.


philipsgithub:
> Looks good! Did you test this on a spot with an arm? If so I'll approve and merge.

hi, I will also add the basic services for the gripper which are already supported by the spot_wrapper. and then test all of them together next week

philipsgithub:
related PR from @heuristicus: https://github.com/bdaiinstitute/spot_wrapper/pull/43

PR #68, opened 2023-06-01
[OC-4] Build a Spot Dance Interface

vgupta-bdai:
Added execute_dance functionality in Spot Driver


spot_driver/spot_driver/spot_ros2.py
```diff
@@ -40,12 +40,14 @@
...
 from spot_msgs.srv import ClearBehaviorFault
 from spot_msgs.srv import SetVelocity
+from spot_msgs.srv import ExecuteDance
 from spot_msgs.srv import GraphNavUploadGraph, GraphNavClearGraph, GraphNavSetLocalization, GraphNavGetLocalizationPose
 
 #####DEBUG/RELEASE: RELATIVE PATH NOT WORKING IN DEBUG
 # Release
 from .ros_helpers import *
 from spot_wrapper.wrapper import SpotWrapper, CameraSource
+from spot_wrapper.spot_dancing import SpotDancing
```
amessing-bdai, line 50: Left a comment in `spot_wrapper`, but is there a reason this needs to  be exposed separately from `SpotWrapper`. As it is just using one more of the various clients from the spot sdk, I would think that it would be contained in the `spot_wrapper` object.

kaiyu-zheng: Same feedback as @amessing-bdai

Guptabot: fixing it to be that way

davidwatkins-bdai (approved):
Do a hardware test and only merge this after `spot_wrapper` #17 is merged.

davidwatkins-bdai:
Yes https://github.com/bdaiinstitute/spot_wrapper/pull/17

PR #69, opened 2023-06-01
[SW-174] Allow camera setting for point cloud streaming; fixing RGB cloud streaming

kaiyu-zheng:
This PR modifies the point_cloud_xyz.launch.py and point_cloud_xyzrgb.launch.py in spot_ros2 to take in a parameter for the camera used for streaming..  It would be straightforward to ‘include’ multiple such launch files if one wants to stream point clouds from different cameras.

This PR also fixes the colored point cloud streaming. To do this, the encoding for JPEG had to be changed to rgb8 because otherwise, the encoding of the image messages is 8UC3 which is not recognized by cv_bridge and causes
```
[component_container-1] terminate called after throwing an instance of 'cv_bridge::Exception'
[component_container-1]   what():  [8UC3] is not a color format. but [rgb8] is. The conversion does not make sense
```

See results in the following screenshots:

(Mono color)
image-20230601-220325

(RGB)
image-20230601-223941

### DIFF
--- a/spot_driver/launch/point_cloud_xyz.launch.py
+++ b/spot_driver/launch/point_cloud_xyz.launch.py
@@ -45,117 +45,35 @@ def generate_launch_description():
     spot_name = LaunchConfiguration('spot_name')
     spot_name_arg = DeclareLaunchArgument('spot_name', description='Name of spot')
 
-    return LaunchDescription([
-        spot_name_arg,
+    camera = LaunchConfiguration('camera')
+    camera_arg = DeclareLaunchArgument('camera', description='Name of camera')
 
-        # launch plugin through rclcpp_components container
-        launch_ros.actions.ComposableNodeContainer(
-            name='frontleft_container',
-            namespace=spot_name,
-            package='rclcpp_components',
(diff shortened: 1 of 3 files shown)


amessing-bdai:
image

erosen-bdai:
I am trying to use this, but when I run the launch file and subscribe to the PointCloud2 topic in RViz, I get no messages and get the following warning:
"[component_container-1] [WARN] [1687538554.359751642] [point_cloud_xyz_frontleft]: New subscription discovered on topic '/Quinn/depth/frontleft/points', requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY"

This happens for both rgb and not rgb versions.

kaiyu-zheng:
Sorry this isn't documented, but you can change QoS settings in RViZ. Borrowing screenshot [from this comment](https://github.com/ros-perception/image_pipeline/pull/814#issuecomment-1549147011)

image

I also got it to work when I set Reliability Policy and Durability to be both "System Default"

PR #70, opened 2023-06-06
FeatureBranch tf frame prefix handling

philipsgithub:
targets same fix as in https://github.com/bdaiinstitute/spot_ros2/pull/62

but also fixes the tf-tree structure during the tf_message creation (setting odom as parent even when prefixed)


philipsgithub:
as of now, my changes here have been fully integrated into https://github.com/bdaiinstitute/spot_ros2/pull/62

philipsgithub:
the only difference is that #62 additionally adds a launchArgument and renames the corresponding variables

jbarry-bdai:
Thanks!  Should this be closed in favor of #62?

philipsgithub:
> Thanks! Should this be closed in favor of #62?

yes sure:)

PR #71, opened 2023-06-06
remove a magic number being multiplied to cmd_duration_secs in handle…

skpawar1305:
…_trajectory

suppose we give a target with duration of 5 seconds, if spot doesn't reach it in 5 seconds, this action continues to run for 51 (5*10.3), fix it

I'm not sure if was intended for something else

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -1560,7 +1560,7 @@ class SpotROS(Node):
                 # check for timeout
                 com_dur = self.get_clock().now() - command_start_time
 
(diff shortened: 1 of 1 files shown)


jbarry-bdai (approved):
We don't use this function but this looks to me very much like a conversion issue gone bad...  @METEORITENMAX any concerns here?

skpawar1305:
> We don't use this function but this looks to me very much like a conversion issue gone bad... @METEORITENMAX any concerns here?

we're performing in Robocup Rescue League now, and we use this function a lot for autonomy. I'll add more changes once we get back to Germany. Then you can take a look

PR #72, opened 2023-06-07
[N/A] Added pre-commit hooks to the repository

davidwatkins-bdai:
This commit makes sure that every commit is checked against `black`, `ruff`, and `mypy`. This also adds type-checking to the entire repository. It contains instructions for installing the `pre-commit` hooks as well as how to contribute to the repository without going through the entire process of verifying. It also will enforce these checks before merges to `main` occur.


amessing-bdai:
@METEORITENMAX We have successfully tested this on our end. Can you test it on your end?

jbarry-bdai (commented):
Can you undo all changes to `conversions.py`?  That's an auto-generated file and shouldn't be edited.

Otherwise this looks good.  Thanks!

PR #74, opened 2023-06-07
use value attribute of wrapper parameter arguments

philipsgithub:
parameter values were not accessed correctly, therefore the wrapper configuration could potentially be unintended


amessing-bdai:
Added these changes to #72 (which will be merged soon) as that is that is easier than merging this in and fixing 72.

PR #78, opened 2023-06-09
[SW-169] Wrapper for RobotCommand and Manipulation action servers

amessing-bdai:
- Delete duplicate single_goal_action_server.py
- Improve shutdown handling
- Use the new single_goal_multiple_action_servers.py

- [x] I have run the BDAII hardware in the loop tests


jbarry-bdai (approved):
Thanks for this!

PR #79, opened 2023-06-09
[N/A] Add all manipulation statuses for handling

amessing-bdai:
- Fixes bug with incorrect goal check for success
- Adds a function to check all feedback status to determine when the action is finished

- [x] I have tested this on BDAII's hardware in the loop tests

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -1333,10 +1333,10 @@ class SpotROS(Node):
         feedback: RobotCommandFeedback = None
         feedback_msg: RobotCommand.Feedback = None
         while (
-                rclpy.ok()
-                and not goal_handle.is_cancel_requested
-                and self._robot_command_goal_complete(feedback) == GoalResponse.IN_PROGRESS
-                and goal_handle.is_active
+            rclpy.ok()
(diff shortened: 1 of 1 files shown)


spot_driver/spot_driver/spot_ros2.py
```diff
@@ -1411,19 +1459,14 @@ def handle_manipulation_command(self, goal_handle: ServerGoalHandle) -> Manipula
...
-
             time.sleep(0.1)  # don't use rate here because we're already in a single thread
 
         # publish a final feedback
         result = Manipulation.Result()
         if feedback is not None:
             goal_handle.publish_feedback(feedback_msg)
             result.result = feedback
-        result.success = self._goal_complete(feedback) == GoalResponse.SUCCESS
+        result.success = self._manipulation_goal_complete(feedback) == GoalResponse.SUCCESS
```
davidwatkins-bdai, line 1469: Should these still be returning this custom `GoalResponse` type?

davidwatkins-bdai (approved):
LGTM

spot_driver/spot_driver/spot_ros2.py
amessing-bdai: I used the same one for consistency

PR #81, opened 2023-06-09
[N/A] try_shutdown for the external shutdown

amessing-bdai:
Got left out when I redid the clean shutdown PR


tkusnur-bdai (approved):
:+1:

PR #82, opened 2023-06-11
[SW-196] Bug fix "follow arm command feedback_choice is 0 'FEEDBACK_NOT_SET'"

kaiyu-zheng:
See details https://theaiinstitute.atlassian.net/browse/SW-196

In short, I noticed that sending a follow arm command leads to error messages like
```
[ERROR] [1686510848.195233146] [Stitch.spot_ros2]: ERROR: unknown mobility command type
```
and as a result, `robot_command_client.send_goal_and_wait(action_goal)` would never return. 

I came up with a fix that makes sense to me and works for my purpose (see videos in the linked ticket). Basically, I believe that when (1) `sync_feedback.mobility_command_feedback_is_set` is True, and (2) `feedback_choice` is actually not set (i.e. set to 0), it indicates a mobility command actually has finished, which means we may return SUCCESS, instead of IN_PROGRESS.

I'm not 100% sure if the fix is correct. But if not, I hope alternative explanations can be provided for the bug I observed.


jbarry-bdai (commented):
Please make sure we understand why this is happening before this change goes in.  My guess is that the error is somewhere else.

amessing-bdai:
In the corresponding [example](https://github.com/boston-dynamics/spot-sdk/blob/master/python/examples/arm_with_body_follow/arm_with_body_follow.py), they simply use `block_until_arm_arrives` ([link](https://github.com/boston-dynamics/spot-sdk/blob/7b0a06c3ae6e23aca03aad85af883761a257c8b1/python/bosdyn-client/src/bosdyn/client/robot_command.py#L1911)), which checks the arm feedback, but doesn't check the mobility feedback. In their various `block` functions they don't check `if *_is_set is True` or use `feedback_choice`. They simply use the `HasField` function and then check specific feedback statuses. I wonder if this is a due to the different in how we check the statuses from how they do?

kaiyu-zheng:
Here is an analysis. Suppose the error is somewhere else, and we'd like to trace down where it is.

Before we start, the mob_feedback proto message upon the error message is:
```
bosdyn_msgs.msg.MobilityCommandFeedback(
   feedback=bosdyn_msgs.msg.MobilityCommandFeedbackOneOfFeedback(
       se2_trajectory_feedback=bosdyn_msgs.msg.SE2TrajectoryCommandFeedback(
           status=bosdyn_msgs.msg.SE2TrajectoryCommandFeedbackStatus(value=0),
           body_movement_status=bosdyn_msgs.msg.SE2TrajectoryCommandFeedbackBodyMovementStatus(value=0)),
       se2_velocity_feedback=bosdyn_msgs.msg.SE2VelocityCommandFeedback(),
       sit_feedback=bosdyn_msgs.msg.SitCommandFeedback(status=bosdyn_msgs.msg.SitCommandFeedbackStatus(value=0)),
       stand_feedback=bosdyn_msgs.msg.StandCommandFeedback(
            status=bosdyn_msgs.msg.StandCommandFeedbackStatus(value=0),
            standing_state=bosdyn_msgs.msg.StandCommandFeedbackStandingState(value=0)),
       stance_feedback=bosdyn_msgs.msg.StanceCommandFeedback(status=bosdyn_msgs.msg.StanceCommandFeedbackStatus(value=0)),
       stop_feedback=bosdyn_msgs.msg.StopCommandFeedback(),
       follow_arm_feedback=bosdyn_msgs.msg.FollowArmCommandFeedback(),
       feedback_choice=0),
   status=bosdyn_msgs.msg.RobotCommandFeedbackStatusStatus(value=1))
```
The error happens because of `feedback_choice=0`, which doesn't match any of the if-checks.

(1) The most suspicious point to start is:
* The error message is thrown under an if-block where `sync_feedback.mobility_command_feedback_is_set == True`.
* But, we also see this a warning message `"WARNING: FollowArmCommand provides no feedback"` inside the if-block.

(2) A natural question is: Why would `sync_feedback.mobility_command_feedback_is_set` be True if FollowArmCommand provides no feedback? 

(3) A reasonable suspicion is, is `sync_feedback.mobility_command_feedback_is_set` incorrectly set to True?  Tracing down where `sync_feedback` comes from...
```python
sync_feedback = feedback.command.synchronized_feedback    # [spot_ros2.py:1158]
feedback = self._get_robot_command_feedback(goal_id)      # [spot_ros2.py:1346]
```
 Inside `._get_robot_command_feedback`, if `goal_id` is not None (as in this case), the feedback is created by
```python
feedback = RobotCommandFeedback()
...
conv.convert_proto_to_bosdyn_msgs_robot_command_feedback(
    self.spot_wrapper.get_robot_command_feedback(goal_id).feedback, feedback
)
```
(4) The part `self.spot_wrapper.get_robot_command_feedback(goal_id)` boils down to just calling a function from SpotSDK
```python
    def get_robot_command_feedback(self, cmd_id):
        return self._robot_command_client.robot_command_feedback(cmd_id)
```
It's reasonable to assume this works correctly.

(5) So the suspicion leads us to `conv.convert_proto_to_bosdyn_msgs_robot_command_feedback` in conversions.py, which traces down as follows
```
convert_proto_to_bosdyn_msgs_robot_command_feedback
-> convert_proto_to_bosdyn_msgs_robot_command_feedback_one_of_command
-> convert_proto_to_bosdyn_msgs_synchronized_command_feedback   (proto.HasField("synchronized_feedback") is True)
-> convert_proto_to_bosdyn_msgs_mobility_command_feedback  (how ros_msg.mobility_command_feedback is set)
-> convert_proto_to_bosdyn_msgs_mobility_command_feedback_one_of_feedback
```
The last function `convert_proto_to_bosdyn_msgs_mobility_command_feedback_one_of_feedback` has a list of if-checks, based on whether `proto.HasField(...)` is True.  If you look at the proto message printed above, all fields are not set. So all the if-checks are skipped. Therefore, `ros_msg.feedback_choice` is not set after `convert_proto_to_bosdyn_msgs_mobility_command_feedback_one_of_feedback` and that causes the error message.

(6) Because the feedback proto message is returned directly by the Spot SDK after the robot has completed the follow-arm command, we should assume that the message itself is correct.  That is, after the robot completes the follow-arm command, we will get a feedback message with none of its fields set. 

(7) So, if indeed something is wrong somewhere else, then `convert_proto_to_bosdyn_msgs_mobility_command_feedback_one_of_feedback` is what's wrong. But, looking at its implementation ([link to code](https://github.com/bdaiinstitute/spot_ros2/blob/5b40d4997d000d82202457af1bc7d8aad5c948a8/spot_driver/spot_driver/conversions.py#L2040)), I can't tell what's wrong.   Sure, one could work around the error reported in this issue by adding an "else" block in `convert_proto_to_bosdyn_msgs_mobility_command_feedback_one_of_feedback` which sets `ros_msg.feedback_choice` to be `ros_msg.FEEDBACK_FOLLOW_ARM_FEEDBACK_SET` , but this is clearly wrong.

(Conclusion) I can't identify the issue somewhere else -- In summary, when the robot completes the follow-arm command, Spot SDK sends a feedback message with none of its fields set.  As a result, it _would be_ the case that the `mob_feedback` object has its `feedback_choice` field to be 0 upon completion of the follow-arm command, and there is nothing wrong with this case. So the fix in this PR basically catches that case and gives it green light.

baxelrod-bdai:
It has been a while since i looked at this code.  but i **think** this is fine.  `mob_feedback.feedback.feedback_choice` certainly has a lot of options.  it is possible one was omitted by accident.  i believe it was my intent to list all the enumerations.  and if I wanted to leave one out, i probably would have commented it.  

and based on the intent of the logic there and how i interpret having feedback unset this sounds fine.  but we should certainly test this a bunch.

kaiyu-zheng:
> It has been a while since i looked at this code. but i **think** this is fine. `mob_feedback.feedback.feedback_choice` certainly has a lot of options. it is possible one was omitted by accident. i believe it was my intent to list all the enumerations. and if I wanted to leave one out, i probably would have commented it.
> 
> and based on the intent of the logic there and how i interpret having feedback unset this sounds fine. but we should certainly test this a bunch.

Just to understand, do you think the fix in this PR is fine?

I tested this on hardware.  You can find the recording when the error occurred in [this video on JIRA](https://theaiinstitute.atlassian.net/browse/SW-196), and [the recording when the fix is applied]((https://theaiinstitute.atlassian.net/browse/SW-196?focusedCommentId=11941)).  Basically, after the fix, the follow arm command test worked as expected.

baxelrod-bdai (approved):

kaiyu-zheng:
Just tested walk to 6d pose and works

PR #86, opened 2023-06-14
Fix setuptools future deprecation in setup.cfg file

zmk5:
When running `colcon` you'll get a warning about a `script-dir` and `install-scripts` will not be supported in future versions and to use the underscored versions. Small fix just to cut down on warnings that pop up during building the package.

### DIFF
--- a/spot_driver/setup.cfg
+++ b/spot_driver/setup.cfg
@@ -1,4 +1,4 @@
 [develop]
-script-dir=$base/lib/spot_driver
+script_dir=$base/lib/spot_driver
 [install]
-install-scripts=$base/lib/spot_driver
+install_scripts=$base/lib/spot_driver


jbarry-bdai:
Have you checked that this still builds and works with `ros2 run`?  If so I'll approve and merge.

zmk5:
> Have you checked that this still builds and works with `ros2 run`? If so I'll approve and merge.

Hey! Yeah I checked using it with a fresh build of `colcon build --symlink-install` and the deprecation warning no longer appears as intended. I ran `ros2 run spot_driver spot_ros2` and it also seems to be working as intended.

Screenshot 2023-06-16 at 11 08 20 AM

jbarry-bdai (approved):

PR #88, opened 2023-06-15
[N/A] Regenerate conversions.py from a sorted list of protos

amessing-bdai:
Previously, the script used to generate the `conversions.py` file would collect the list of proto files in the spot-sdk in an arbitrary order due to using `os.walk`. This makes it challenging to generate the same file when checking for regression. This new version of `conversions.py` was generated by sorting the list of proto files after the `os.walk` before they are processed. This should allow future bug fixes and spot-sdk upgrades to be easier to check for regression.

To be clear, this new file has no changes except a reordering of the functions as a result of processing the same proto files in a different order


davidwatkins-bdai (approved):
Approved under the assumptions you ran the appropriate tests to make sure this didn't break anything. Good work.

PR #90, opened 2023-06-20
Create Spot CORE I/O Extension

rshu-bd:
# Summary
This PR packages the Spot driver into a Spot Extension so it can be run on a CORE I/O. The PR only includes build files, not the actual `.spx` extension file. This has to be built locally. 

[Extensions](https://dev.bostondynamics.com/docs/payload/docker_containers) are software packages or static files introduced in 3.2 that can be installed or simply uploaded to Boston Dynamics CORE I/O or Scout platforms. Once the Extension is running additional ROS Nodes can be run from another machine connected to the robot's WiFi to send `RobotCommands` and visualize robot data.

The benefits of running the Spot ROS 2 Driver as an Extension on the CORE I/O are:

- Run ROS on the robot.
- ROS Spot Driver will continue to run even after comms are lost with a personal computer.
- Offload computation from your machine to the CORE I/O.

# Details
- The Docker container is running Ubuntu 22.04 and ROS Humble. 
- Built a version of [bosdyn_msg arm64 deb package](https://git.bostondynamics.com/BostonDynamics/rt/blob/a8c1f8ae1775e793a4ed115e9ab0a477b91b48a3/apps/coreio/extensions/ros2_driver/ros-humble-bosdyn-msgs_0.0.0-0jammy_arm64.deb) to target the container's OS and ROS version. It would be good to add this as a release to `bosdyn_msgs` repo
- Add authentication with payload credentials in addition to using environment variables or ROS params to define user credentials. 
   - [PR#27](https://github.com/bdaiinstitute/spot_wrapper/pull/27) adds authentication with payload credential to the `spot_wrapper`


ggoretkin-bdai:
Very cool!

I think I am misunderstanding the claim:
> Offload computation from your machine to the CORE I/O.

Only one Spot extension can be loaded at a time, so if  i

ggoretkin-bdai:
very sorry, unintentional closing!

jbarry-bdai:
@baxelrod-bdai @brennan-vanderlaan-bd-ai thoughts on who might be a good person from the SWE/devops side to look at this PR?

spot_coreio_extension/images/monitoring_spot_ros_extension.gif
brennan-vanderlaan-bd-ai, line 1: Does this binary file need to exist in the repo or could it be in Notion with supporting documentation?

spot_coreio_extension/update_bashrc
```diff
@@ -0,0 +1,5 @@
+# Adding all the necessary ros sourcing
+echo "" >> ~/.bashrc
+echo "## ROS" >> ~/.bashrc
+echo "source /opt/ros/$ROS_DISTRO/setup.bash" >> ~/.bashrc
+echo "source $SPOT_WS/install/setup.bash" >> ~/.bashrc
```
baxelrod-bdai, line 5: maybe this is a naive question about Docker.  But what if the install is run twice?  Then these lines will be put in the `.bashrc` twice.  can we put some smarts in the script to grep for these lines and skip adding them again if they are already in there?

spot_coreio_extension/manifest.json
```diff
@@ -0,0 +1,6 @@
+{
+  "description": "Spot ROS 2 Driver",
+  "version": "3.2.0",
+  "icon": "icon.png",
```
baxelrod-bdai, line 4: what is this icon for?  is it really needed?

spot_coreio_extension/images/monitoring_spot_ros_extension.gif
baxelrod-bdai: I agree, there seem to be lots of images as documentation in this PR that might be better put elsewhere.

spot_coreio_extension/update_bashrc
rshu-bd: By install are you referring to building the docker image, installing the extension on the CORE I/O or something different? 

When building the docker image this will only get copied once, but `source /opt/ros/$ROS_DISTRO/setup.bash` could be a duplicate if the base image already has this in the `.bashrc` 

The installation in the CORE I/O will load the docker image and run containers of that image. These steps will not modify the `.bashrc` of the docker image. So no issue of duplication there. 

However, I think this may be redundant. The docker image also defines the entry point which sources the same directories. Will test if this is true. 
https://github.com/bdaiinstitute/spot_ros2/blob/b067b7c0d21b2fb874cd1b7ab28dd7ff29264727/spot_coreio_extension/Dockerfile.arm64#L58

rshu-bd: Delete the `update_basrch` step not required.

spot_coreio_extension/images/monitoring_spot_ros_extension.gif
rshu-bd: It doesn't need to be in this repo. Can adapt to follow your documentation practices.

jbarry-bdai:
Just an update: we're working on testing this with our Core I/Os!  @ggoretkin-bdai @mpickett-bdai @bhung-bdai 

@rshu-bd if you get a chance to fix the conflict it might make for a better test but not the end of the world

rshu-bd:
> Just an update: we're working on testing this with our Core I/Os! @ggoretkin-bdai @mpickett-bdai @bhung-bdai
> 
> @rshu-bd if you get a chance to fix the conflict it might make for a better test but not the end of the world

Great thank you for the update. Yes, I can work through the conflicts

jbarry-bdai:
@bibit-bdai @mkeck-bdaii this PR could be a solution to the latency issues while logging because it runs the ROS2 driver on the Core I/O.  @ggoretkin-bdai and @bhung-bdai are working on getting it running for us if you're interested

spot_coreio_extension/README.md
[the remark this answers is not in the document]
rshu-bd: > The web interface throw an internal server error when I attempt an extension upload (unresolved, bypassed by copying the docker to the Core I/O and running it from SSH)

Would need more information and screenshots to debug this.

>The authentication fails when you direct the core i/o to use the payload_guid (solved by setting password and env variables) 

Payload authentication won't work until [PR #27](https://github.com/bdaiinstitute/spot_wrapper/pull/27) on spot_wrapper is merged.


> Setting up the FastDDS on the core i/o causes the ROS topics to be sent to a different port than spot_ros2 is expecting on the Core I/O, meaning the examples will not run locally (resolved by temporarily stopping the FastDDS setup)

We use the FastDDS server to route ROS Discovery messages through the open ports in the CoreI/O so that a computer connected to the robot's access point can also see the ROS Nodes and Topics. It seems like the session in which you try to launch the spot_driver is not using the FastDDS Discovery Server, did you export `ROS_DISCOVERY_SERVER=192.168.80.3:21000`  env variable? Also, are you using docker-compose or launching the containers individually? 

> Running spot_ros2 examples on the Core I/O is failing due to the service claim not being triggered (unresolved)

I'm assuming this happens after you stopped the Discovery Server? 

> To clarify, was the setup ever intended to run ROS packages directly on the Core I/O? Or does this particular setup require being running ROS packages solely on the offboard machine? Is the Core I/O meant to have access to the ROS topics?

You should be able to run ROS packages on the Core I/O and/or offboard. This particular setup runs the `spot_driver` on the Core I/O

> In addition, was this setup tested outside of directly connecting in access point mode?

This setup was **only** tested connecting in access point mode.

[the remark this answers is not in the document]
bhung-bdai: > Would need more information and screenshots to debug this.

I was informed that it may be a timeout or an upload size issue. Unfortunately, the only information I can provide is that the upload finishes and during the processing it gives this: 

coreio_upload_web_issues

> Payload authentication won't work until [PR #27](https://github.com/bdaiinstitute/spot_wrapper/pull/27) on spot_wrapper is merged.

Got it, thanks!

> We use the FastDDS server to route ROS Discovery messages through the open ports in the CoreI/O so that a computer connected to the robot's access point can also see the ROS Nodes and Topics. It seems like the session in which you try to launch the spot_driver is not using the FastDDS Discovery Server, did you export `ROS_DISCOVERY_SERVER=192.168.80.3:21000` env variable? Also, are you using docker-compose or launching the containers individually?

I am using the docker-compose.yml with docker-compose up. I have exported that variable in several test attempts, as well as other IP addresses between both Wi-Fi access and direct access point connection. None of those have worked and ROS2 topic list on both the Core I/O and remote machine cannot see the list of topics. I'm not sure if that will be resolved by your "Known Issues" section and setting everything to a SuperClient. 
 
> I'm assuming this happens after you stopped the Discovery Server?

This was an error in our spot_ros2 example code.

> You should be able to run ROS packages on the Core I/O and/or offboard. This particular setup runs the `spot_driver` on the Core I/O

Got it. Spot_ros2 runs locally on the Core I/O without the FastDDS. The tricky part has been getting the ROS information publishing to the correct location, which is likely an error in how I set up FastDDS or resolvable by your suggestions on fixing the ROS CLI. Do you have further suggestions?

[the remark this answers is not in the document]
bhung-bdai: I was able to resolve most of the issues on my local setup. The current issue that I am running into is if there a way to avoid needing to add the IP route to access the Core I/O through Spot? If we want to access multiple local Core I/Os and control several spots on the same network, that doesn't seem sustainable

rshu-bd:
> My $0.02: we should not use the web interface to upload container images. We would like a deployment solution that is scriptable, so using `scp` and the instructions at https://dev.bostondynamics.com/docs/payload/docker_containers#install-extension-using-a-usb-drive , as @bhung-bdai is doing, is the better solution for us. I do not think it is worth debugging the web interface issue for our use case.

There are multiple methods of uploading a Spot Extension to the CoreI/O. If one works better than the other for your workflow that is great. However, all methods should work so it may be a symptom of some other underlying issue.

bhung-bdai:
> > My $0.02: we should not use the web interface to upload container images. We would like a deployment solution that is scriptable, so using `scp` and the instructions at https://dev.bostondynamics.com/docs/payload/docker_containers#install-extension-using-a-usb-drive , as @bhung-bdai is doing, is the better solution for us. I do not think it is worth debugging the web interface issue for our use case.
> 
> There are multiple methods of uploading a Spot Extension to the CoreI/O. If one works better than the other for your workflow that is great. However, all methods should work so it may be a symptom of some other underlying issue. It would be good to find out what the issue is.

Thanks for the timely replies Roberto. If you can, could you look into the networking comment I made above? We are hoping to avoid ip routing through Spot as it will be inconvenient if we have multiple spots.

spot_coreio_extension/README.md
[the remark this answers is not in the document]
rshu-bd: Looking into it. We did not test running multiple Spots.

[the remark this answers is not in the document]
bhung-bdai: Currently the way I've been getting the ROS messages from the Core I/O over the WiFi (not direct access mode) has been by using the route you added 
`sudo ip route 192.168.50.5 via <spot ip>`
My concern is that if we want to have two spots running, then we cannot add 192.168.50.5 for spot 1 and spot 2 or else there will be conflicts.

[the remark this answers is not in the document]
notero-bd: When ROS nodes run on the Core IO they advertise themselves using the Core IO's IP address. Unfortunately all the Core IO's have the same IP address, so if you have two ROS nodes on two different Core IOs, they will advertise themselves as having the same IP address.

I don't know of a way to solve this. The solution would have to involve ROS nodes advertising themselves as having Spot's IP address, and let the forwarding rules do their job. I'll do some research to see if DDS supports this, but I'm not optimistic it will.

ggoretkin-bdai:
I'll attempt to clarify the networking situation

In general, there are multiple Spots, each with a Core I/O. We won't be connected to each of their WAPs, or have all the Spot robots be on `192.168.50.3`. Ultimately we need a solution that works with [the network configuration provided by BD](https://support.bostondynamics.com/s/article/Payload-device-network-configuration), where both the Spot and the Core I/O are visible to the "building" network under the same IP address, with port forwarding. So any steps that add routes regarding `192.168.*.*` to our workstations/laptops are not suitable for our use case.

notero-bd:
> I'll attempt to clarify the networking situation
> 
> In general, there are multiple Spots, each with a Core I/O. We won't be connected to each of their WAPs, or have all the Spot robots be on `192.168.50.3`. Ultimately we need a solution that works with [the network configuration provided by BD](https://support.bostondynamics.com/s/article/Payload-device-network-configuration), where both the Spot and the Core I/O are visible to the "building" network under the same IP address, with only some ports forwarded to the Core I/O. So any steps that add routes regarding `192.168.*.*` to our workstations/laptops are not suitable for our use case.
> 
> [EDIT] Just saw @notero-bd 's reply here: [#90 (comment)](https://github.com/bdaiinstitute/spot_ros2/pull/90#discussion_r1261457844) . Thanks! I think we are on the same page.

Is the only node running on the Core IO the spot wrapper? Do you need dynamic discovery of other nodes on the Core IO?

One solution would be using DEFAULT_FASTRTPS_PROFILES.xml to hardcode the IP address of each spot wrapper node to the IP address of the corresponding Spot. I've never tried this and I'm not 100% sure it would work. We can try it out. But if you need dynamic discovery of nodes on the Core IO this won't work.

notero-bd:
A possible solution is GRE tunneling, sample instructions here: https://unix.stackexchange.com/questions/404084/how-to-configure-a-gre-tunnel-for-ubuntu

notero-bd:
There is another, possibly easier and cleaner, solution that is worth trying out. FastDDS does support advertising endpoints with IP addresses different than the ones available through local interfaces. They are called [external locators](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/announced_locators.html). If this works as I understand it to, we could eliminate the VPN and not have to worry about tunneling.

I won't have cycles to try this out until the end of next week, if you guys want to give it a try before then, let me know how it goes.

bhung-bdai:
> There is another, possibly easier and cleaner, solution that is worth trying out. FastDDS does support advertising endpoints with IP addresses different than the ones available through local interfaces. They are called [external locators](https://fast-dds.docs.eprosima.com/en/latest/fastdds/transport/announced_locators.html). If this works as I understand it to, we could eliminate the VPN and not have to worry about tunneling.
> 
> I won't have cycles to try this out until the end of next week, if you guys want to give it a try before then, let me know how it goes.

Thank you for the notice Nick. I have found some examples and can test it out.

jbarry-bdai:
@jschornak-bdai i'm tagging you here in case you have a magical DDS solution for this!

jschornak-bdai:
> @jschornak-bdai i'm tagging you here in case you have a magical DDS solution for this!

No magical solution by the looks of it, unfortunately. I'll need to learn more about the specifics of the Core I/O system and Spot's payload device network config to understand how we could massage DDS messages across the network.

At PickNik we use CycloneDDS as our primary ROS 2 middleware, and the way we approach configuring multimachine systems is to create a `cyclonedds_config.xml` file for each computer that explicitly lists the IP addresses of its peers and the network interface to use. This allows discovering nodes on other systems without requiring multicast, but of course the downside is if you want to use your laptop as a debugging platform to introspect ROS messages you need to add it as a peer to all the other systems.

rshu-bd:
@ggoretkin-bdai @bhung-bdai @notero-bd What work remains to merge this PR? I'm aware there is still some work to get multiple robots running on the same network. I'm wondering if we can merge this PR and in a later PR add support for multiple Spots?

bhung-bdai:
> @ggoretkin-bdai @bhung-bdai @notero-bd What work remains to merge this PR? I'm aware there is still some work to get multiple robots running on the same network. I'm wondering if we can merge this PR and in a later PR add support for multiple Spots?

Hi Roberto,
I'm not sure it's a great idea to merge it now. The main concerns I have are the following:

- The deployment process is not working over the extensions portal. I have documentation to add about deploying over `ssh` which isn't in this README.md
- The can of worms which is ROS2. I've found that the solution of ip route for a single robot requires a bit more of a careful setup on the remote client and the Core I/O environment variable setup. While this is the only way we can get ROS2 CLI to work right now, I believe the solution isn't correctly calibrated to how ROS2 should be framed. I  fully believe that regardless of one vs many machines, the merged solution should be one of the following:

1. Setting fastdds_external_locators with ROS2 humble, which I have gotten to work on my machine but am having trouble installing on Docker. This works with n-many robots and is the correct way to set ROS2 communications on Humble with this network setup. However, it doesn't work with ROS2 CLI tools which is a huge detriment becaues you cannot configure a discovery server using external locators in the current state of `fastdds`. 
2.  A fully detailed setup for the Core I/O and local machine to networks as peer to peer VPN clients. This is a very involved process and many will not be happy to have so much overhead.
3. Upgrade to Iron due to the [subnet multicast discovery](https://docs.ros.org/en/iron/Releases/Release-Iron-Irwini.html#improved-discovery-options). This will require building our driver for a new system and testing if it even works. I have no guarantees it does.

This requires a lot of extra work to ensure it works smoothly.

rshu-bd:
@bhung-bdai Thank you for the update and explanation. 

> The deployment process is not working over the extensions portal. I have documentation to add about deploying over ssh which isn't in this README.md

This should not be the case and we should fix it. I'll reach out offline to discuss this further. 

> The can of worms which is ROS2. I've found that the solution of ip route for a single robot requires a bit more of a careful setup on the remote client and the Core I/O environment variable setup. While this is the only way we can get ROS2 CLI to work right now, I believe the solution isn't correctly calibrated to how ROS2 should be framed. 

Got it. It seems like none of the solutions are perfect and all of them have some blocker. Is there one out of these three that you are leaning towards?

bhung-bdai:
> @bhung-bdai Thank you for the update and explanation.
> 
> > The deployment process is not working over the extensions portal. I have documentation to add about deploying over ssh which isn't in this README.md
> 
> This should not be the case and we should fix it. I'll reach out offline to discuss this further.
> 
> > The can of worms which is ROS2. I've found that the solution of ip route for a single robot requires a bit more of a careful setup on the remote client and the Core I/O environment variable setup. While this is the only way we can get ROS2 CLI to work right now, I believe the solution isn't correctly calibrated to how ROS2 should be framed.
> 
> Got it. It seems like none of the solutions are perfect and all of them have some blocker. Is there one out of these three that you are leaning towards?

The ideal long term solution would probably be to upgrade ROS2 if the DDS changes work as intended. I can't imagine them rolling back such a useful change.

bhung-bdai:
> > @bhung-bdai Thank you for the update and explanation.
> > > The deployment process is not working over the extensions portal. I have documentation to add about deploying over ssh which isn't in this README.md
> > 
> > 
> > This should not be the case and we should fix it. I'll reach out offline to discuss this further.
> > > The can of worms which is ROS2. I've found that the solution of ip route for a single robot requires a bit more of a careful setup on the remote client and the Core I/O environment variable setup. While this is the only way we can get ROS2 CLI to work right now, I believe the solution isn't correctly calibrated to how ROS2 should be framed.
> > 
> > 
> > Got it. It seems like none of the solutions are perfect and all of them have some blocker. Is there one out of these three that you are leaning towards?
> 
> The ideal long term solution would probably be to upgrade ROS2 if the DDS changes work as intended. I can't imagine them rolling back such a useful change.

Given the issues with the Core I/O integrating into our system and the fact it doesn't meet our original intended needs for image processing speed after some testing, we've decided it's not a priority to develop at the moment and are holding off indefinitely.

baxelrod-bdai (changes requested):
we should close this PR

rshu-bd:
Closing PR. We can reopen if there is new interest or improvements in the current network/communication issues

Issue #33, opened 2023-04-03
Dock/undock and other services?


julianraheema:
In spot_ros recently added the dock/undock and other services. Is there a plan to implement these services in spot_ros2? Also in the spot_ros, we can launch `roslaunch spot_vis view_robot.launch` to interactive with robot, and use the services? I am not sure how to do that with spot_ros2.

jbarry-bdai:
We're actually in the process of combining the back end for this with the ROS1 wrapper so it will be easier to add those services.  We don't have interactive markers or anything set up in ROS2 yet.

julianraheema:
@jbarry-bdai I appreciate your feedback, and please point me to the new stuff on ROS2 once are ready. I think it will be a good idea to migrate these services and features from ros1 to ros2. The sooner the better :) and Thank you for your great work!

jbarry-bdai:
That just went in!  @heuristicus can you point us at the ROS1 code?  Then I can look at how it needs to change for ROS2 and put up a few instructions :)

heuristicus:
It's basically all of the stuff in [spot_viz](https://github.com/heuristicus/spot_ros/tree/master/spot_viz). There are some rviz configurations for panel layouts, with corresponding launch files to view the robot model (useful when checking your urdf extensions), and view_robot will show you some basic output when the driver is running.

The `resource` directory contains a `.ui` file which is loaded by the spot panel when it starts. This file is made using Qt designer and saves a ton of messing around with GUI stuff directly in the code (and there is still a lot of messing around).

The panel itself is in the src directory. I don't know how plugins work in ros2, but in ros1 it is necessary to have a `plugin_description.xml` which is referred to in the package xml with

```
  <export>
      <rviz plugin="${prefix}/plugin_description.xml"/>
  </export>
```

The panel itself just interacts directly with the services and topics made available by the robot.

julianraheema:
Thank you very much @jbarry-bdai and @heuristicus I will try to try to integrate the spot_viz to ros2, but there is something that I need to take care of using ROS1 first which is the navigation stack. If you guys know a good reference for the navigation stack for SPOT, I would appreciate it. I was able to create a map using rtabmap, and I think I am now at the point of integrating the navigation stack. Once I am done with this then I can go back and migrate to ROS2. Again thank you for your support and great work.

zmk5:
[...]

PR #91, opened 2023-06-21
Spot Docking and Undocking Services

zmk5:
This pull request will add services for both docking and undocking. The former gets its own ROS Service message to allow a user to specify a dock ID, while undocking is a `Trigger`. I originally wrote this to have the dock services always run but instead opted to use a config file to supply a `has_dock` parameter which will start the services when needed. A user can modify the `has_dock` parameter in the example `yaml` to enable use of the dock.

I've tested this with a mock behavior tree code I've written on a physical spot robot and have run and passed all the pre-commit checks. This pull request should address issue #33.


heuristicus:
My preference for these kinds of things is for them to be available, unless the system physically does not have the capability. For example, if the robot is a non-enterprise model, perhaps disabling the service would make sense, but otherwise I'd have it available - the user can decide whether or not to use it, and the standard error messages from the system should suffice if it is called for some reason.

zmk5:
That makes sense. I was under the assumption that the dock was an extra and doesn't come with Spot, so I made it an option.

If the maintainers want, I can make the service always available with a new commit.

jbarry-bdai (commented):
I agree with @heuristicus that I think this should always be available rather than having people set it in the config file.  If you want to make that change and verify you tested docking, I'll approve and merge!

One other thought: things like "Dock" that last for a long time are really somewhat better as actions rather than services so that they can be interrupted.  That said, it's a lot more work to add an action than a service and this adds useful functionality so I don't object to putting it in as is.

Thanks for making a contribution!

zmk5:
> I agree with @heuristicus that I think this should always be available rather than having people set it in the config file. If you want to make that change and verify you tested docking, I'll approve and merge!
> 
> One other thought: things like "Dock" that last for a long time are really somewhat better as actions rather than services so that they can be interrupted. That said, it's a lot more work to add an action than a service and this adds useful functionality so I don't object to putting it in as is.

Understood! I removed the `had_dock` parameter and verified it works with Spot 3.3.0.

> 
> Thanks for making a contribution!

No problem! We were adding this functionality and thought to support the team while we work on our stuff.

jbarry-bdai (approved):

PR #95, opened 2023-06-27
[SW-232] Fixed conversions for map values which was submsgs from the protobuf

amessing-bdai:
This fixes a bug from the proto2ros conversion. 

protobufs allow submessages e.g.
```
message Outside
{
  message Inside
  {
    ...
  }
}
```

whereas ros msgs do not. When we generate the ros msgs we concatenate the message names, so the inner message would make a ros msg file for `OutsideInside`. protobuf also allows for `map`s whereas ros msgs do not, so we create a `Key<key-type>Value<value-type>` ros msg file. Previous proto2ros was not using the concatenated name for the import when converting a type that contained a map. It now does and this is the result. This specifically is need to use some of the manipulation API


davidwatkins-bdai:
Is this automatically generated or is this custom? Have you tested this change?

amessing-bdai:
Yes, it is autogenerated and yes

jbarry-bdai (approved):
Oh yeah @wrangel-bdai noticed this on an earlier PR.  Thanks for the fix!

davidwatkins-bdai (approved):

PR #98, opened 2023-06-28
[N/A] Increased default rate to be the same as the largest individual task rate and set to float

amessing-bdai:
The default robot state task rate is 50 Hz (in code at least) but gets throttled because the async_tasks_rate is lower. This increases the default to match the largest individual callback rate. It also changes it to explicitly be a float, so a float can be used for the ros parameter.

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -204,7 +204,7 @@ class SpotROS(Node):
 
         self.declare_parameter("deadzone", 0.05)
         self.declare_parameter("estop_timeout", 9.0)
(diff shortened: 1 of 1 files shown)


jbarry-bdai:
I see the change to the number but nothing else?

Do we want to add logic to check if the async task rate is lower than the largest task rate and print a warning or increase it to that rate?

amessing-bdai:
> I see the change to the number but nothing else?
> 
> Do we want to add logic to check if the async task rate is lower than the largest task rate and print a warning or increase it to that rate?

I think I will print a warning and increase it

davidwatkins-bdai (approved):
Nice catch on this

PR #105, opened 2023-07-11
SW-276-fix-the-ros-walk-forward-example

bhung-bdai:
The namespace for the robot wasn't included, causing an error where it wouldn't find the ROS topic it was supposed to publish to. I've added it to the example as a command line argument to pass in and updated the README

### DIFF
--- a/examples/simple_walk_forward/README.md
+++ b/examples/simple_walk_forward/README.md
@@ -19,7 +19,7 @@ ros2 launch spot_driver spot_driver.launch.py
 
 5.  Run the example:
 ```bash
-ros2 run simple_walk_forward walk_forward
+ros2 run simple_walk_forward walk_forward --robot <spot_name>
 ```
 
 The robot should walk forward.
--- a/examples/simple_walk_forward/README.md
+++ b/examples/simple_walk_forward/README.md
@@ -16,12 +16,19 @@ This is a simple example of using ROS2 to make the robot walk 1m forward.
 ```bash
 ros2 launch spot_driver spot_driver.launch.py
 ```
+If you want to launch with a namespace,
+```bash
+ros2 launch spot_driver_with_namespace.launch.py spot_name:=<spot_name> 
+```
 
 5.  Run the example:
 ```bash
+ros2 run simple_walk_forward walk_forward
+```
+If you are launching spot_ros2 with a namespace, use the following command instead:
+```bash
 ros2 run simple_walk_forward walk_forward --robot <spot_name>
 ```
-
 The robot should walk forward.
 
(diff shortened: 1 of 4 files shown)


jbarry-bdai (commented):
Actually this one is supposed to be very simple and work without a robot name... was it broken if you follow the README instructions?  Many places don't use the name prefix because they only have one Spot so I've been trying to preserve that functionality.

bhung-bdai:
> Actually this one is supposed to be very simple and work without a robot name... was it broken if you follow the README instructions? Many places don't use the name prefix because they only have one Spot so I've been trying to preserve that functionality.

Not exactly. The issue happens when bdai spot_driver launches, since that provides a namespace and it accordingly renames the ROS transforms. I can avoid launching that way if that's the intention but you can always provide no robot name. I just figured that most of us are launching with namespace, so it makes more sense to provide it as an option

bhung-bdai:
Should we leave it the way it is or optionally pass in a name?

jbarry-bdai:
> Should we leave it the way it is or optionally pass in a name?

I went back and forth on that and decided not to have that option because it made it a little more complicated and I was trying to keep it as simple as possible because it's an example.  What would be really good though would be to add another example showing how to use the robot prefix and how to visualize that in rviz and everything.  I wanted to do that but didn't get around to it...

bhung-bdai:
I'll try nixing this change and doing a quick double check if it works 100%
following the readme tomorrow then. I was having some issues on the core-io
without namespace l but they might have been resolved in my install.

Brandon

jbarry-bdai:
I would definitely try this on a laptop before using the core i/o.  it's possible it is broken!  we're not as good as we should be about running these examples with change to the driver.

bhung-bdai:
Reopening this because it's a small change and improves quality of life if you need to test different setups. Now both ROS examples work with and without a namespace

examples/simple_walk_forward/README.md
```diff
@@ -19,7 +19,7 @@ ros2 launch spot_driver spot_driver.launch.py
 
 5.  Run the example:
 ```bash
-ros2 run simple_walk_forward walk_forward
+ros2 run simple_walk_forward walk_forward --robot <spot_name>
```
amessing-bdai, line 22: Since the example now works with or without a robot name that should be explained here.

amessing-bdai: Should also be done in the readme for simple_arm_motion

amessing-bdai (changes requested):

amessing-bdai (approved):

PR #112, opened 2023-07-26
[SW-328] Better handling of arm0.link_wr1 for hand camera transforms

ksharma-bdai:
## Overview
This patch rolls back the changes made in #107 (see description for problem notes) in favor of a cleaner and simpler approach to handling `arm0.link_wr1` in hand camera image callbacks - with special case code to:

- avoid publishing a static transform for `arm0.link_wr1` at all as it is not static
- rename parent `arm0.link_wr1` to `link_wr1` in any static transforms for proper connectivity to robot state updates
- rely on robot state publication to handle all necessary dynamic tfs in a single place

This change will have no impact on spot configurations without an arm and will not break anything in the future if the naming discrepancy is fixed by Boston Dynamics.

Relying on robot state has the added benefit of much higher frequency tf updates, eliminating the choppy behavior seen when relying on only image callback timing.

In addition, this elimates `TF_OLD_DATA` warnings that were observed due to the currently serial nature of our image pipeline callbacks. For example:

- a batch of depth image callbacks for some timestamp window [0, n] arrive
- each callback publishes its tf snapshot at reported capture timestamp
- a batch of rgb image callbacks arrive next for an overlapping window [0, n + 1]
- any attempts to publish within the overlapping window (- some delta) are refused with warning, they're too old

## Testing
updated behavior, smooth updates at robot state (see #107 for older behavior)

https://github.com/bdaiinstitute/spot_ros2/assets/137219344/f522e4b9-c6b0-44b1-b437-cbef129dac38

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -2101,12 +2101,33 @@ class SpotROS(Node):
             frame_prefix = self.spot_wrapper.frame_prefix
         excluded_frames = [self.tf_name_vision_odom.value, self.tf_name_kinematic_odom.value, frame_prefix + "body"]
         excluded_frames = [f[f.rfind("/") + 1 :] for f in excluded_frames]
+
+        # Special case handling for hand camera frames that reference the link "arm0.link_wr1" in their
+        # transform snapshots. This name only appears in hand camera transform snapshots and appears to
+        # be a bug in this particular image callback path.
+        #
+        # 1. We exclude publishing a static transform from arm0.link_wr1 -> body here because it depends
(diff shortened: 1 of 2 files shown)


spot_driver/spot_driver/spot_ros2.py
```diff
@@ -922,7 +922,7 @@ def publish_camera_images_callback(self, image_type: SpotImageType) -> None:
             image_info_pub = getattr(self, f"{image_entry.camera_name}_{publisher_name}_info_pub")
```
ksharma-bdai, line 922: all changes to `spot_ros2.py` are roll backs for #107

jbarry-bdai (approved):
Thanks!

PR #114, opened 2023-07-27
[SW-324] Fix incorrect list_world_objects endpoint call

ksharma-bdai:
This patch corrects a minor bug where the `list_world_objects` callback was attempting to invoke a non-existent function/endpoint on the `spot_wrapper` rather than one of its members.


jbarry-bdai (approved):
@bmay-bdai this may have been your issue with the dock stuff...

amessing-bdai:
@jbarry-bdai It was. This is the resulting fix for that

PR #116, opened 2023-07-28
Spot tour guide functionalities

fengjun-yang1:
Added functionalities:
- service MovePTZ: set the pan, tilt, zoom (ptz) of the payload Spot CAM
- service GetPTZ: get current pan, tile, zoom of the Spot CAM
- action NavigateToDynamic: navigation with dynamic speed

Added parameter in spot_ros.yaml:
- bool enable_payload_cam: we added this because we realized that decoding the incoming image stream from the payload cam is compute intensive. Without this variable, the image streaming will start automatically. This variable serves as a flag to turn off image streaming from the payload camera.


myeatman-bdai (changes requested):
I would rename everything that references PTZ to mention something like "SpotCamPTZ". Looks good otherwise.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -355,9 +363,13 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None) -> N
             if not self.spot_wrapper.is_valid:
                 return
 
-            try:
-                self.spot_cam_wrapper = SpotCamWrapper(self.ip, self.username, self.password, self.cam_logger)
-            except SystemError:
+            enable_payload_cam: Parameter = self.get_parameter("enable_payload_cam").value
```
amessing-bdai, line 366: Can this parameter be added to `spot_driver/config/spot_ros_example.yaml`?

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -355,9 +363,13 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None) -> N
...
+            if enable_payload_cam:
+                try:
```
amessing-bdai, line 368: Is the try/except a check for whether the robot has a spot cam? If so, can a comment be added that explains this? Also, a warning should probably be logged that says the user enabled the spot cam, but the robot does not have one.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2090,6 +2174,58 @@ def handle_navigate_to(self, goal_handle: ServerGoalHandle) -> NavigateTo.Result
...
+            time.sleep(1)
+
+    def handle_navigate_to_dynamic(self, goal_handle: ServerGoalHandle) -> NavigateToDynamic.Result:
+        """Navigate to a given goal waypoint. Unlike navigate_to,
+        this method supports velocity adjustment and pausing/cancellation during a navigation request
+        (see handle_set_navigate_to_params)"""
+
+        self.goal_handle_dynamic = goal_handle
+        # Create thread to periodically publish feedback
+        feedback_thread = threading.Thread(target=self.handle_navigate_to_dynamic_feedback, args=())
```
amessing-bdai, line 2199: Is there a way to do this without creating a separate thread for feedback? There are already some threading bottlenecks that we are working on improving

amessing-bdai (changes requested):

PR #122, opened 2023-08-09
[SW-341] Fix list_graph after graphnav modularization update

ksharma-bdai:
Fix `list_graph` after the recent `spot_wrapper` modularization - it no longer takes an (unused) argument for the graph - which must be uploaded prior. This patch also removes some excessive error level logging that would appear as errors when everything was working as intended.


bhung-bdai (approved):
LGTM. Didn't realize the graphnav module didn't have an option to just load the graph from the file given.

PR #136, opened 2023-08-24
[SW-440] Fixed the goals by passing a name to them

bhung-bdai:
The examples would fail originally due to the ActionClientWrapper refactor requiring an action name and a goal. The previous examples only sent a goal. Added an action name for the `walk_forward` and `arm_simple` examples.


amessing-bdai (approved):

amessing-bdai:
As in the name of the Action type? The ros ActionClient doesn't have a name. We use this name for logging, so it wouldn't be very useful if all robot commands just said "Sending a RobotCommand"

bhung-bdai:
> Hm I'm slightly confused why we don't at least have that action name default to the action name for the client but given that it is currently this way (@amessing-bdai ?) this seems like the right thing to do!

I will go change that in addition to changing this. I think it just got overlooked because we don't actually use it anywhere internally.

amessing-bdai:
@bhung-bdai I am not convinced we should do that

PR #139, opened 2023-09-06
[N/A] Remove foxy from README as we don't actually test for it anymore

amessing-bdai:


bhung-bdai (approved):
LGTM. Have we decided on how to handle backward version compatibility in the future?

PR #125, opened 2023-08-14
Update Spot URDF Xacro and rviz launch parameter

abaker-bdai:
This PR address [PN-31](https://theaiinstitute.atlassian.net/browse/PN-31) and [PN-67](https://theaiinstitute.atlassian.net/browse/PN-67)

This PR provides the following updates to the `spot_description` and `spot_driver` packages:
- Converted main spot and spot arm urdfs into xacro macros in order to add parameters.
- added parameters to include arm in urdf, define prefixes for joint/link names, optionally specify a parent frame.
- renamed `spot.urdf.xacro` to `spot_macro.xacro`
- renamed `spot_with_arm.urdf.xacro` to `spot.urdf.xacro` and refactored to use `spot_macro.xacro`
- updated launch files to use `spot_macro.xacro` 
- removed namespaces launch file as it is no longer needed.

Testing:
You can test the updates by launch the rviz with the robot model by running the following command:
``` bash
ros2 launch spot_description description.launch.py arm:=true prefix:=test_
```
If you wish to test using the spot driver, run the following command:
``` bash
ros2 launch spot_driver spot_driver.launch.py arm:=true prefix:=test_
```

[PN-31]: https://theaiinstitute.atlassian.net/browse/PN-31?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PN-67]: https://theaiinstitute.atlassian.net/browse/PN-67?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #141, opened 2023-09-11
Fix launch file param for RViz config filepath

abaker-bdai:
This is a follow up PR to #125.
This PR changes the `rviz_config_filename` parameter to `rviz_config_file` and doesn't attempt to look for the rviz config inside of the `spot_driver` package. 

Also includes documentation changes that remove the reference to `spot_driver_with_namespace.launch.py`


bhung-bdai (approved):
LGTM

PR #143, opened 2023-09-12
[PN-79] Add Depth Registration component nodes to spot_driver launch file

abaker-bdai:
This PR adds an additional launch file for launch Depth Registration component nodes. This launch file is then used inside of spot_driver.launch.py. 

A user can specify a list of depth cameras to launch by passing the path to a yaml file containing a list to the parameter `camera_sources_yaml`.

Testing:
To test on Opal, run the following command:
``` bash
ros2 launch spot_driver spot_driver.launch.py has_arm:=true launch_rviz:=true spot_name:=Opal camera_sources_yaml:=/workspaces/bdai/ws/src/external/spot_ros2/spot_driver/config/camera_sources_example.yaml
```

### DIFF
--- a/spot_driver/launch/spot_depth_publishers.launch.py
+++ b/spot_driver/launch/spot_depth_publishers.launch.py
@@ -7,14 +7,14 @@ from launch.substitutions import LaunchConfiguration, PathJoinSubstitution
 
 
 def launch_depth_register_nodelets(
-    context: launch.LaunchContext, spot_name: LaunchConfiguration, ld: launch.LaunchDescription
+    context: launch.LaunchContext,
+    spot_name: LaunchConfiguration,
+    camera_sources: LaunchConfiguration,
+    ld: launch.LaunchDescription,
 ) -> None:
-    # TODO: provide these names as LaunchConfiguration inputs to allow higher-level functions to set them
-    camera_sources = ["frontleft", "frontright", "left", "right", "back", "hand"]
-
     composable_node_descriptions = []
 
-    for camera in camera_sources:
+    for camera in camera_sources.perform(context):
         composable_node_descriptions.append(
             launch_ros.descriptions.ComposableNode(
                 package="depth_image_proc",
@@ -56,9 +56,14 @@ def generate_launch_description() -> launch.LaunchDescription:
     spot_name = LaunchConfiguration("spot_name")
     spot_name_arg = DeclareLaunchArgument("spot_name", description="Name of spot")
 
-    # ... other code omitted for clarity ...
+    camera_sources = LaunchConfiguration("camera_sources")
+    camera_sources_arg = DeclareLaunchArgument(
+        "camera_sources",
+        default_value=["frontleft", "frontright", "left", "right", "back", "hand"],
+        description="List of camera sources",
+    )
 
-    ld = launch.LaunchDescription([spot_name_arg, ...])
+    ld = launch.LaunchDescription([spot_name_arg, camera_sources_arg])
 
-    ld.add_action(OpaqueFunction(function=launch_depth_register_nodelets, args=[spot_name, ld]))
+    ld.add_action(OpaqueFunction(function=launch_depth_register_nodelets, args=[spot_name, camera_sources, ld]))
     return ld
--- a/spot_driver/launch/spot_depth_publishers.launch.py
+++ b/spot_driver/launch/spot_depth_publishers.launch.py
@@ -1,0 +1,64 @@
+# Copyright [2023] Boston Dynamics AI Institute, Inc.
+
+import launch
+import launch_ros
+from launch.actions import DeclareLaunchArgument, OpaqueFunction
+from launch.substitutions import LaunchConfiguration, PathJoinSubstitution
+
+
+def launch_depth_register_nodelets(
+    context: launch.LaunchContext, spot_name: LaunchConfiguration, ld: launch.LaunchDescription
+) -> None:
+    # TODO: provide these names as LaunchConfiguration inputs to allow higher-level functions to set them
+    camera_sources = ["frontleft", "frontright", "left", "right", "back", "hand"]
+
+    composable_node_descriptions = []
+
+    for camera in camera_sources:
+        composable_node_descriptions.append(
+            launch_ros.descriptions.ComposableNode(
+                package="depth_image_proc",
+                plugin="depth_image_proc::RegisterNode",
+                name="register_node_" + camera,
+                namespace=spot_name,
+                # Each entry in the remappings list is a tuple.
+                # The first element in the tuple is the internal name of the topic used within the nodelet.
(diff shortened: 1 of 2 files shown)


spot_driver/launch/spot_depth_publishers.launch.py
```diff
@@ -0,0 +1,80 @@
...
+    composable_node_descriptions = []
+
+    camera_sources_yaml = camera_sources_yaml.perform(context)
+    if not camera_sources_yaml or camera_sources_yaml == "None":
+        camera_sources = ["frontleft", "frontright", "left", "right", "back", "hand"]
+    else:
+        with open(camera_sources_yaml, "r") as yaml_file:
+            camera_sources = yaml.safe_load(yaml_file)["sources"]
+
+    cameras_str = ""
```
jschornak-bdai, line 26: What's this used for?

spot_driver/launch/spot_depth_publishers.launch.py
```diff
@@ -0,0 +1,80 @@
...
+    ld: launch.LaunchDescription,
+) -> None:
+    composable_node_descriptions = []
+
+    camera_sources_yaml = camera_sources_yaml.perform(context)
+    if not camera_sources_yaml or camera_sources_yaml == "None":
+        camera_sources = ["frontleft", "frontright", "left", "right", "back", "hand"]
+    else:
+        with open(camera_sources_yaml, "r") as yaml_file:
+            camera_sources = yaml.safe_load(yaml_file)["sources"]
```
jschornak-bdai, line 24: From the perspective of feature parity with the existing driver I think it's not actually necessary to allow users to specify which exact cameras should have registered depth image nodelets.

Within the existing spot_ros2 driver, there are essentially two factors which determine which registered depth images are published:
- The `"publish_depth_registered"` parameter: If true, request and publish registered depth images for all Spot cameras. If false, no registered depth images are published.
- Whether the Spot has an arm or not: If it does, publish all requested image types for the hand camera. If it does not, no hand camera images are published. (Note that while the driver launch files do have a "has_arm" argument, within the driver node it determines if the Spot has an arm by actually querying the robot through the SDK.)

This means that currently the option to publish registered depth images (or images of any type, for that matter) is an "all or nothing" sort of thing. This is the case [in the standalone image publisher node](https://github.com/bdaiinstitute/spot_ros2/blob/c945b72b21898284015409261c1095786471f076/spot_driver/spot_driver/publish_cameras.py#L129-L131) as well. I think it would be OK to just always create a nodelet for every expected Spot camera, and add a launch configuration argument to set whether or not we expect the hand camera to be available as well. This would simplify the configuration of your launch file quite a bit, since you wouldn't need to handle parsing the camera sources as an input YAML file.

spot_driver/launch/spot_driver.launch.py
```diff
@@ -106,6 +109,14 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
 
     ld.add_action(rviz)
 
+    depth_images = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(
+            [PathJoinSubstitution([FindPackageShare(THIS_PACKAGE), "launch", "spot_depth_publishers.launch.py"])]
+        ),
+        launch_arguments={"spot_name": spot_name, "camera_sources_yaml": camera_sources_yaml}.items(),
+    )
+    ld.add_action(depth_images)
```
jschornak-bdai, line 118: Could you add a launch argument to toggle between using the new separate registered depth nodelets and using the previous approach of requesting registered depth images through the Spot SDK? It should default to `true`, and if it's `true` it should disable the parameter in the spot_ros2 driver node to request registered depth images. If it's `false` the opposite should happen -- the new nodelets should not be brought up, and if the `publish_depth_registered` parameter is `true` it should be passed in to the spot_ros2 driver node as usual.

spot_driver/launch/spot_driver.launch.py
```diff
@@ -106,6 +109,14 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
 
     ld.add_action(rviz)
 
+    depth_images = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(
+            [PathJoinSubstitution([FindPackageShare(THIS_PACKAGE), "launch", "spot_depth_publishers.launch.py"])]
```
jschornak-bdai, line 114: Make sure you're consistent about how you refer to these nodelets, since it's important that they're specifically publishing registered depth images. I'd suggest using more detailed names for the launch file and the `depth_images` variable here.

jschornak-bdai (changes requested):

spot_driver/launch/spot_depth_publishers.launch.py
abaker-bdai: It was used for debugging, I'll remove this

spot_driver/launch/spot_depth_publishers.launch.py
abaker-bdai: Noted, I'll make that change

jschornak-bdai (approved):
Looks good! Note that I won't have a chance to test this out on hardware until next Monday, so someone who has done that should review this as well.

spot_driver/launch/spot_depth_registered_publishers.launch.py
```diff
@@ -0,0 +1,70 @@
...
+def launch_depth_register_nodelets(
+    context: launch.LaunchContext,
+    spot_name: LaunchConfiguration,
+    has_arm: LaunchConfiguration,
+    ld: launch.LaunchDescription,
+) -> None:
+    composable_node_descriptions = []
+
+    camera_sources = ["frontleft", "frontright", "left", "right", "back"]
+    if has_arm.perform(context):
```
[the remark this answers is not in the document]
abaker-bdai: Should be fixed now

spot_driver/launch/spot_driver.launch.py
```diff
@@ -106,6 +113,19 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
...
+    registered_depth_image_nodes = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(
+            [
+                PathJoinSubstitution(
+                    [FindPackageShare(THIS_PACKAGE), "launch", "spot_depth_registered_publishers.launch.py"]
+                )
+            ]
+        ),
+        launch_arguments={"spot_name": spot_name, "has_arm": has_arm}.items(),
+        condition=IfCondition(use_depth_registered_nodelets),
```
bhung-bdai, line 125: If this is check on if `use_depth_registered_nodelets == "False" or ues_depth_registered_nodeltes == "false"`, it would nice to use this for `has_arm` instead of the current workaround.

bhung-bdai (approved):
LGTM. One cosmetic question about how something works but the change works as intended.

PR #145, opened 2023-09-20
[PN-83][PN-84] Add C++ Spot image publisher node

jschornak-bdai:
## Description

~~Depends on PN-80 to integrate the Spot C++ SDK and its dependencies into this repo. CI will fail on this PR until that work is completed.~~

- Adds `SpotImagePublisherNode`, which requests image data from Spot at a rate of 15 Hz and publishes the images it receives to ROS topics.
- Launches the C++-based Spot image publisher node in the main Spot driver launch file.
- Disables the Python driver node's image publishing functionality, since this is replaced by the new image publisher node.

### Authenticate using Parameters

Try connecting to Spot by providing the address and credentials as ROS parameters. Replace the placeholder values in the command below with actual values.

```
ros2 run spot_driver_cpp spot_image_publisher_node --ros-args -p address:=YOUR_SPOT_ADDRESS -p username:=YOUR_USERNAME -p password:=YOUR_PASSWORD
```

The node should launch successfully, and images from Spot's RGB and depth cameras should be published on topics in the global namespace.

### Authenticate using Environment Variables

The default approach for passing in auth information in the existing spot_driver nodes is to use three environment variables: `SPOT_IP`, `BOSDYN_CLIENT_USERNAME`, and `BOSDYN_CLIENT_PASSWORD`. Set values for these variables in your environment, and then run the node using the following command:

```
ros2 run spot_driver_cpp spot_image_publisher_node
```

The node should launch successfully, and images from Spot's RGB and depth cameras should be published on topics in the global namespace.

The frame IDs published to `/tf_static` should be named like `left_fisheye` (note that there should be no leading `/`).

### Setting Spot's Name Using a Namespace

The new image publisher uses the namespace of the ROS node as the name of Spot. Provide a namespace like below:

```
ros2 run spot_driver_cpp spot_image_publisher_node --ros-args -r __ns:=/MySpotName
```

The image topics should now be named like `/MySpotName/camera/frontright/image`.

The frame IDs published to `/tf_static` should be named like `MySpotName/left_fisheye` (note that there should be no leading `/`).

### Configure Published Image Types Using Parameters

The types of image that will be published can be configured through ROS parameters, like this:

```
ros2 run spot_driver_cpp spot_image_publisher_node --ros-args -p publish_rgb:=true -p publish_depth:=true -p publish_depth_registered:=true
```

Toggling the boolean values between `true` and `false` when launching the node should make the corresponding topics appear and disappear.

### Launch using the driver launch file

In an environment where you have exported values for SPOT_IP`, `BOSDYN_CLIENT_USERNAME`, and `BOSDYN_CLIENT_PASSWORD`, launch the driver nodes:

```
ros2 launch spot_driver spot_name:=MySpotName tf_prefix:=MySpotName
```

This should bring up both the Python driver node that controls Spot and publishes dynamic robot state data and the C++ image publisher node that requests data from Spot's cameras.

Only one publisher should be publishing on each image topic.


spot_driver_cpp/src/interfaces/spot_interface.cpp
```diff
@@ -0,0 +1,335 @@
...
+  else if (nanos_unskewed >= 1e9)
+  {
+    nanos_unskewed -= 1e9;
+    seconds_unskewed += 1;
+  }
+
+  // If the timestamp contains a negative time, create an all-zero ROS Time.
+  if (seconds_unskewed < 0)
+  {
+    return builtin_interfaces::build<builtin_interfaces::msg::Time>().sec(0).nanosec(0);
```
mpickett-bdai, line 80: Can I get a definition on the clock_skew? And does this specific case happen because the spot api time is running before the ROS driver?

jschornak-bdai: The Boston Dynamics Spot SDK docs have a section about the clock sync service: https://dev.bostondynamics.com/docs/concepts/base_services#time-sync

I'll add a doc comment describing what this does more comprehensively.

This specific line is to handle an edge case if the Spot API time is running ahead of the driver PC's system clock AND the epoch is very small. In practice I wouldn't expect this to happen, but it's a possibility due to the differences between ROS and Spot SDK time types.

mpickett-bdai: Yeah a little comment/documentation just say which time reference we are going from -> to would be helpful here. Also if the Spot time is always running ahead would this mean the ROS Time would constantly be behind or do we reset the spot time somewhere to sink them?

spot_driver_cpp/src/interfaces/spot_interface.cpp
```diff
@@ -0,0 +1,335 @@
...
+}
+
+tl::expected<sensor_msgs::msg::Image, std::string> toImageMsg(const bosdyn::api::ImageCapture& image_capture, const std::string& robot_name, const google::protobuf::Duration& clock_skew)
+{
+      const auto& image = image_capture.image();
+      auto data = image.data();
+
+      std_msgs::msg::Header header;
+      header.frame_id = robot_name + "/" + image_capture.frame_name_image_sensor();
+      header.stamp = applyClockSkew(image_capture.acquisition_time(), clock_skew);
```
mpickett-bdai, line 154: Is the reasoning for not using the convertRobotTimeToLocalTime function here because you want to use a specific clock_skew?

spot_driver_cpp/include/spot_driver_cpp/interfaces/rclcpp_wall_timer_interface.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <rclcpp/timer.hpp>
+#include <spot_driver_cpp/interfaces/timer_interface_base.hpp>
+
+#include <chrono>
+#include <memory>
+
+namespace spot_ros2
+{
+/**
+ * @brief Implements TimerInterfaceBase to use rclcpp's WallTimer.
```
jbarry-bdai, line 15:
```suggestion
 * @brief Implements Spot's TimerInterfaceBase to use rclcpp's WallTimer.
```

spot_driver_cpp/include/spot_driver_cpp/spot_image_sources.hpp
```diff
@@ -0,0 +1,29 @@
...
+ * @brief Create a list of image sources corresponding to the specified image types.
+ * 
+ * @param get_rgb_images Sets whether to request RGB images.
+ * @param get_depth_images Sets whether to request depth images.
+ * @param get_depth_registered_images Sets whether to request registered depth images.
+ * @param has_hand_camera Sets whether to request images from the hand camera.
+ * @return ImageSources 
+ */
+std::vector<ImageSource> createImageSourcesList(const bool get_rgb_images, const bool get_depth_images, const bool get_depth_registered_images, const bool has_hand_camera);
+}
```
jbarry-bdai, line 29: Namespace closing brackets should end with `}  // <namespace name>`

spot_driver_cpp/launch/spot_image_publisher.launch.xml
```diff
@@ -0,0 +1,28 @@
...
+  <arg name="username" />
+  <arg name="password" />
+
+  <!-- Args for optional parameters: -->
+  <arg name="spot_name" default="" />
+  <arg name="image_quality" default="70.0"/>
+  <arg name="rgb_cameras" default="true"/>
+  <arg name="publish_rgb" default="true"/>
+  <arg name="publish_depth" default="true"/>
+  <arg name="publish_depth_registered" default="false"/>
```
jbarry-bdai, line 13: I think we should work to defaulting to publishing rgb and depth for all cameras and doing depth registered calculation on this side and publishing that.  I think that's inline with these defaults so just making a note :)

spot_driver_cpp/launch/spot_image_publisher.launch.xml
```diff
@@ -0,0 +1,28 @@
...
+
+  <node pkg="spot_driver_cpp" exec="spot_image_publisher_node" namespace="$(var spot_name)" output="screen">
+    <!-- Note: Spot's name is used as the namespace for the node, rather than being passed in as a parameter.
+         This simplifies publishing topics with the correct prefix, since the name will be used as the prefix. -->
```
jbarry-bdai, line 17: If spot's name is empty then this should be in the global namespace

spot_driver_cpp/launch/spot_image_publisher.launch.xml
```diff
@@ -0,0 +1,28 @@
+<launch>
+  <!-- Args for required parameters: -->
+  <arg name="address" />
+  <arg name="username" />
+  <arg name="password" />
```
jbarry-bdai, line 5: Are these required?  What happens if they're specified in environment parameters?

spot_driver_cpp/src/interfaces/rclcpp_tf_interface.cpp
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  tl::expected<void, std::string> RclcppTfInterface::publishStaticTransforms(const std::vector<geometry_msgs::msg::TransformStamped>& transforms)
+  {
+    // Keep track of which transforms the static transform publisher is already publishing.
+    std::vector<geometry_msgs::msg::TransformStamped> new_transforms;
+    for (const auto& transform : transforms)
+    {
+      const auto frame_id_pair = std::make_pair(transform.header.frame_id, transform.child_frame_id);
+      // Do not publish transforms which have already previously been published.
```
jbarry-bdai, line 19: I don't think that's right.  I think the static transform broadcaster only remembers the last message.  So you have to publish all the transforms you want broadcast even if they've been previously broadcast.  Static transform broadcasts are costly though so you only want to publish a new one if it's a new transform is my understanding.

spot_driver_cpp/src/interfaces/spot_interface.cpp
```diff
@@ -0,0 +1,394 @@
...
+  // We exclude the odometry frames from static transforms since they are not static. We can ignore the body
+  // frame because it is a child of odom or vision depending on the preferred_odom_frame, and will be published
+  // by the non-static transform publishing that is done by the state callback
+  "body",
+  "odom",
+  "vision",
+
+  // Special case handling for hand camera frames that reference the link "arm0.link_wr1" in their
+  // transform snapshots. This name only appears in hand camera transform snapshots and appears to
+  // be a bug in this particular image callback path.
```
jbarry-bdai, line 39:
```suggestion
  // transform snapshots. This name only appears in hand camera transform snapshots and is a known
  // bug in the Spot API.
```

spot_driver_cpp/src/interfaces/spot_interface.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+  const auto authenticate_result = robot_->Authenticate(username, password);
+  if(!authenticate_result)
+  {
+    return tl::make_unexpected("Authentication with provided username and password did not succeed.");
+  }
+
+  // Start time synchronization between the robot and the client system.
+  // This must be done only after a successful authentication.
+  const auto start_time_sync_response = robot_->StartTimeSync();
```
jbarry-bdai, line 269: Does this block until time sync is established?  If not we want to do that before returning.

spot_driver_cpp/src/interfaces/spot_interface.cpp
```diff
@@ -0,0 +1,394 @@
...
+  GetImagesResult out;
+  for (const auto& image_response : get_image_result.response.image_responses())
+  {      
+      const auto& image = image_response.shot().image();
+      auto data = image.data();
+
+      const auto image_msg = toImageMsg(image_response.shot(), robot_name_, clock_skew_result.value());
+      if (!image_msg)
+      {
+        std::cerr << "Failed to convert SDK image response to ROS Image message: " << image_msg.error() << std::endl;
```
jbarry-bdai, line 333: Do we want to pass in a logger instead of going to cerr?  This would allow us to pass in the ROS logger and have these messages go in logs and stuff without a ROS dependency.

spot_driver_cpp/src/interfaces/spot_interface.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+      const auto info_msg = toCameraInfoMsg(image_response, robot_name_, clock_skew_result.value());
+      if (!info_msg)
+      {
+        std::cerr << "Failed to convert SDK image response to ROS CameraInfo message: " << info_msg.error() << std::endl;
+        continue;
+      }
+
+      const auto& camera_name = image_response.source().name();
+      if(const auto result = fromSpotImageSourceName(camera_name); result.has_value())
```
jbarry-bdai, line 345:
```suggestion
      if (const auto result = fromSpotImageSourceName(camera_name); result.has_value())
```

jbarry-bdai: I might give up the scoping for readability here and put the result assignment before the if statement (same comment applies in multiple places)

spot_driver_cpp/src/spot_image_publisher.cpp
```diff
@@ -0,0 +1,144 @@
...
+            image_request->set_image_format(get_raw_rgb_images ? bosdyn::api::Image_Format_FORMAT_RAW : bosdyn::api::Image_Format_FORMAT_JPEG);
+        }
+        else if (source.type == SpotImageType::DEPTH)
+        {
+            bosdyn::api::ImageRequest* image_request = request_message.add_image_requests();
+            image_request->set_image_source_name(source_name);
+            image_request->set_quality_percent(kDefaultDepthImageQuality);
+            image_request->set_image_format(bosdyn::api::Image_Format_FORMAT_RAW);
+        }
+        else // SpotImageType::DEPTH_REGISTERED
```
jbarry-bdai, line 52:
```suggestion
        else  // SpotImageType::DEPTH_REGISTERED
```
Two spaces between code and comments

spot_driver_cpp/src/spot_image_sources.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+const std::vector<std::string> kImageSourceNames = {
+    "back",
+    "frontleft",
+    "frontright",
+    "left",
+    "right",
+};
+
+constexpr auto kImageSourceHand = "hand";
```
jbarry-bdai, line 17: Are these names not defined in the Spot C++ api anywhere?

jbarry-bdai (changes requested):
Exciting to see this coming together!

Block is for the static transform handling.

One overall style note: Braces go on the same line as the function declaration / conditional / loop.

Some tests we should make sure we do:

- [x] Empty robot name
- [x] Non-empty robot name
- [x] Spot specifics (username, password, IP) set via environment variables
- [x] Spot specifics set via parameters
- [ ] Check that static camera frames are all being broadcast even if they come in sequentially
- [x] Check that the hand camera shows up in the correct place (this is the `arm0.link_wr1` bug)
- [x] Check that robot clock skew is applied correctly
- [ ] Check that the camera info messages published by this C++ driver are identical to the ones published by the Python driver

jschornak-bdai:
> One overall style note: Braces go on the same line as the function declaration / conditional / loop.

For this and the other style/formatting comments, I'll add clang-format to the precommit config and run it on all the new files. Do you already have a config in another project that reflects the BDAII style guide?

spot_driver_cpp/src/interfaces/spot_interface.cpp
jschornak-bdai: > It should only get set to zero if the timestamp is smaller than the clock skew and the timestamp is epoch time right? So that should basically never happen? Just checking my understanding.

This is my understanding as well -- I would only expect this to happen if both Spot and the host PC think it's very early in the morning on January 1, 1970.

spot_driver_cpp/src/interfaces/spot_interface.cpp
jschornak-bdai: Right, we only need to get one clock skew for the current image request, and we can reuse it for each image, camera_info, and transform message we publish here.

spot_driver_cpp/src/interfaces/spot_interface.cpp
jschornak-bdai: I can make these return as the "unexpected" result from this function, which will cause the messages to be logged to the user in SpotImagePublisher. I was sort of on the fence about doing that, but since the reasons these failure can happen  a )are not spurious/intermittent, and b) are indicative of either an implementation bug or something that requires corrective action by the user, I think it's more useful to stop publishing images and return the error immediately.

jbarry-bdai:
> > One overall style note: Braces go on the same line as the function declaration / conditional / loop.
> 
> For this and the other style/formatting comments, I'll add clang-format to the precommit config and run it on all the new files. Do you already have a config in another project that reflects the BDAII style guide?

@llee-bdai @ksharma-bdai @baxelrod-bdai do we?

amessing-bdai:
From the main repo
```
-   repo: https://github.com/ssciwr/clang-format-hook.git
    rev: v12.0.1  # Use the sha / tag you want to point at
    hooks:
    -   id: clang-format
        types_or: [c++, c, cuda]
-   repo: https://github.com/cpplint/cpplint.git
    rev: 1.6.1
    hooks:
    -   id: cpplint
        args: ['--quiet']
        exclude_types: [cuda]
```

spot_driver_cpp/src/spot_image_sources.cpp
jschornak-bdai: The fully-qualified names of the camera sources are defined [here in the SDK library](https://github.com/bdaiinstitute/spot-cpp-sdk/blob/master/cpp/bosdyn/client/image/image_source_names.h), but these prefixes aren't included there. It's useful to have them defined to let us parse the SDK's source names into our own types, especially since the SDK doesn't include functions to convert them the other way.

spot_driver_cpp/src/interfaces/rclcpp_tf_interface.cpp
jschornak-bdai: After digging into this a bit it looks like [the rclcpp version](https://github.com/ros2/geometry2/blob/f2e7e1417ea599e719adbf08cd386fdca78e82ec/tf2_ros/src/static_transform_broadcaster.cpp) of the StaticTransformBroadcaster functions differently than [the rclpy version](https://github.com/ros2/geometry2/blob/rolling/tf2_ros_py/tf2_ros/static_transform_broadcaster.py).

In [the rclcpp version's sendTransform function](https://github.com/ros2/geometry2/blob/f2e7e1417ea599e719adbf08cd386fdca78e82ec/tf2_ros/src/static_transform_broadcaster.cpp#L52-L72), it has some logic to compare the new transform against previously-published transforms to determine if the stored list of previously-published transforms needs to be updated, and it always publishes all previously-published transforms when sendTransform is called.

I added a doc comment explaining what's going on behind the scenes here, and I updated the logic in the interface class to only call sendTransform if there's an entirely new transform to avoid unnecessarily broadcasting redundant static transforms.

jbarry-bdai: Where is the exploding head emojii in github reviews

spot_driver_cpp/src/interfaces/spot_interface.cpp
jschornak-bdai: Based on my understanding of how this client works in the Spot SDK, it does block until the time sync is established. In combination with then next request to `GetTimeSyncThread` a few lines later, the clock sync will be ready to use before this function returns.

spot_driver_cpp/src/interfaces/rclcpp_tf_interface.cpp
```diff
@@ -0,0 +1,33 @@
...
+      // If one of the transforms is to a new child frame, flag that a new transform needs to be published and add the
+      // child frame to the set of currently-published frames.
+      if (current_static_child_frames_.count(transform.child_frame_id) == 0)
+      {
+        has_new_frame = true;
+        current_static_child_frames_.insert(current_static_child_frames_.end(), transform.child_frame_id);
+      }
+    }
+
+    // Only publish if there is a new transform.
```
jbarry-bdai, line 26:
```suggestion
    // Only publish if there is a new transform (note that unlike in Python the static transform publisher will correctly re-publish all previous transforms as well).
```

spot_driver_cpp/include/spot_driver_cpp/interfaces/parameter_interface_base.hpp
```diff
@@ -0,0 +1,42 @@
...
+protected:
+  // These are the definitions of the default values for optional parameters.
+  static constexpr auto kDefaultAddress = "10.0.0.3";
+  static constexpr auto kDefaultUsername = "user";
+  static constexpr auto kDefaultPassword = "password";
+  static constexpr double kDefaultRGBImageQuality{ 70.0 };
+  static constexpr bool kDefaultHasRGBCameras{ true };
+  static constexpr bool kDefaultPublishRGBImages{ true };
+  static constexpr bool kDefaultPublishDepthImages{ true };
+  static constexpr bool kDefaultPublishDepthRegisteredImages{ true };
```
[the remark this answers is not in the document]
jschornak-bdai: Depends on if we want this to function identically to the Python spot_ros2 driver or have it work differently. Originally I had these as non-optional, where initialization would fail if the user did not provide a value for the address, username, or password, but I changed it to how it is currently to match the multi-level fallback strategy the Python driver uses.

For the time being I think it makes sense to tend towards feature parity, and we can come back in and change it in a future PR.

spot_driver_cpp/test/test_spot_image_publisher.cpp
```diff
@@ -0,0 +1,372 @@
...
+using ::testing::Property;
+using ::testing::Return;
+using ::testing::HasSubstr;
+using ::testing::_;
+
+namespace spot_ros2::testing
+{
+constexpr auto kExampleAddress{ "192.168.0.10"};
+constexpr auto kExampleUsername {"spot_user" };
+constexpr auto kExamplePassword { "hunter2" };
```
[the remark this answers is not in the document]
jschornak-bdai: Formatting should be fixed now that I ran clang-format on the whole diff.

spot_driver_cpp/src/interfaces/spot_interface.cpp
```diff
@@ -0,0 +1,394 @@
...
+      if (!pixel_format_cv)
+      {
+        return tl::make_unexpected("Failed to determine pixel format: " + pixel_format_cv.error());
+      }
+
+      if (image.format() == bosdyn::api::Image_Format_FORMAT_JPEG)
+      {
+        // When the image is JPEG-compressed, it is represented as a 1 x (width * height) row of bytes.
+        // First we create a cv::Mat which contains the compressed image data...
+        const cv::Mat img_compressed{ 1, image.rows() * image.cols(), CV_8UC1, &data.front()};
```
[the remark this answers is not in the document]
jschornak-bdai: This would be a good thing to investigate and improve as part of a future PR. I think as currently implemented this is adequately performant (especially compared to the performance of the current Python-based image publisher, oof), but there's probably an opportunity to squeeze some extra efficiency out of this too.

spot_driver_cpp/src/spot_image_publisher.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+#include <iostream>
+#include <memory>
+#include <optional>
+#include <stdexcept>
+
+namespace
+{
+constexpr auto kImageCallbackPeriod = std::chrono::duration<double>{1.0 / 15.0 };  // 15 Hz
+constexpr auto kDefaultDepthImageQuality = 100.0;
```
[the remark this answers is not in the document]
jschornak-bdai: As far as I understand the `quality` setting has no effect for depth images, since Spot does not allow setting them to use JPEG compression. I'd need to test with a robot to see if we can omit this entirely, though.

spot_driver_cpp/src/spot_image_sources.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+    if (image_source.type == SpotImageType::RGB)
+    {
+        return std::string("camera").append("/").append(image_source.name);
+    }
+    else if (image_source.type == SpotImageType::DEPTH)
+    {
+        return std::string("depth").append("/").append(image_source.name);
+    }
+    else  // SpotImageType::DEPTH_REGISTERED
```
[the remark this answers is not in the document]
jschornak-bdai: I think the only benefit from doing that would be if Spot gained support for an additional built-in camera, which would require updating the enums and conversion functions anyway. Seems unnecessary in my opinion, especially since I wouldn't actually be able to write a test case to trigger the default failure case given that we handle all possible values of each enums.

spot_driver_cpp/src/interfaces/rclcpp_parameter_interface.cpp
```diff
@@ -0,0 +1,138 @@
...
+constexpr auto kEnvVarNamePassword = "BOSDYN_CLIENT_PASSWORD";
+
+constexpr auto kParameterNameAddress = "address";
+constexpr auto kParameterNameUsername = "username";
+constexpr auto kParameterNamePassword = "password";
+constexpr auto kParameterNameRGBImageQuality = "image_quality";
+constexpr auto kParameterNameHasRGBCameras = "rgb_cameras";
+constexpr auto kParameterNamePublishRGBImages = "publish_rgb";
+constexpr auto kParameterNamePublishDepthImages = "publish_depth";
+constexpr auto kParameterNamePublishDepthRegisteredImages = "publish_depth_registered";
```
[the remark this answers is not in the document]
jschornak-bdai: These are specifically used for declaring ROS 2 parameters, so I think it makes sense to keep them with the rclcpp-specific implementation of the class. I'd contrast these with the default values of the parameters, which I did put in the base class because we'd want to use them across different implementations of the class (such as the test fixtures, for example).

spot_driver_cpp/include/spot_driver_cpp/interfaces/rclcpp_wall_timer_interface.hpp
jschornak-bdai: Strictly speaking we're implementing our own `TimerInterfaceBase` class here rather than a part of the Spot SDK, so I think it's more correct to keep this comment written as-is.

spot_driver_cpp/src/interfaces/rclcpp_publisher_interface.cpp
```diff
@@ -0,0 +1,51 @@
...
+    info_publishers_.clear();
+
+    for (const auto& image_source : image_sources)
+    {
+        // Since these topic names do not have a leading `/` character, they will be published within the namespace of the node, which should match the name of the robot.
+        // For example, the topic for the front left RGB camera will ultimately appear as `/MyRobotName/camera/frontleft/image`.
+        const auto topic_name_base = toRosTopic(image_source);
+
+        const auto image_topic_name = topic_name_base + "/image";
+        image_publishers_.try_emplace(image_topic_name, node_->create_publisher<sensor_msgs::msg::Image>(image_topic_name, rclcpp::QoS(1)));
```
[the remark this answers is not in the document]
jschornak-bdai: There should only be a single ImageSource for a given combination of camera type and image type, so there should be exactly one possible entry into this map for each ImageSource.

To do a better job of enforcing this, I changed how the collection of ImageSources is generated and passed around so it's stored in a `std::set` instead of a `std::vector`. This more clearly communicates our expectation that there cannot be duplicate ImageSources.

spot_driver_cpp/launch/spot_image_publisher.launch.xml
jschornak-bdai: I added env var parsing to set the default values for these. This launch file is an alternative entrypoint for how this new node will ultimately be run in `spot_driver.launch.py` (see the follow-on work in #162 for how that will look), so these launch arguments won't be used in that case.

spot_driver_cpp/src/spot_image_publisher.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include <spot_driver_cpp/types.hpp>
+
+#include <iostream>
+#include <memory>
+#include <optional>
+#include <stdexcept>
+
+namespace
+{
+constexpr auto kImageCallbackPeriod = std::chrono::duration<double>{1.0 / 15.0 };  // 15 Hz
```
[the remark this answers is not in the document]
jschornak-bdai: I think the main case for a per-camera rate setting would be to get the hand camera publishing at its maximum-possible rate of 30Hz. We can address that in follow-up work if we'd like.

install_spot_ros2.sh
```diff
@@ -20,7 +20,8 @@ else
 fi
 
 # Install ROS dependencies
-sudo apt install -y ros-$ROS_DISTRO-joint-state-publisher-gui ros-$ROS_DISTRO-tf-transformations ros-$ROS_DISTRO-xacro ros-$ROS_DISTRO-depth-image-proc
+# TODO(jschornak-bdai): use rosdep to install these packages by parsing dependencies listed in package.xml
```
bhung-bdai, line 23: I assume this is a WIP, so I'm just going to leave a sticky note here.

spot_driver_cpp/src/interfaces/rclcpp_publisher_interface.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+void RclcppPublisherInterface::createPublishers(const std::set<ImageSource>& image_sources) {
+  image_publishers_.clear();
+  info_publishers_.clear();
+
+  for (const auto& image_source : image_sources) {
+    // Since these topic names do not have a leading `/` character, they will be published within the namespace of the
+    // node, which should match the name of the robot. For example, the topic for the front left RGB camera will
+    // ultimately appear as `/MyRobotName/camera/frontleft/image`.
+    const auto topic_name_base = toRosTopic(image_source);
```
bhung-bdai, line 26: Is this function getting the namespace of the robot? It doesn't appear to be referencing it below.

bhung-bdai (commented):
Just a few comments from me. The other thing I wanted to know is how this plays with the composable nodelets that @abaker-bdai had created.

install_spot_ros2.sh
jschornak-bdai: Let's address this one in a different PR -- the current version of this follows the same pattern as installing the previous dependencies, while the TODO proposes changing how dependency installation works in a deeper way.

spot_driver_cpp/src/interfaces/rclcpp_publisher_interface.cpp
jschornak-bdai: The namespace of the node is the same as the namespace of the robot, so because the publisher is publishing its topic within the namespace of the node it handles prepending the robot name to the topic name. We don't need to explicitly set the robot namespace within this function as a result of that.

bhung-bdai (changes requested):
This was addressed in a Slack thread with Joe, but it appears the hand node is not being recorded and published over at the moment.

spot_driver_cpp/src/interfaces/spot_interface.cpp
```diff
@@ -0,0 +1,394 @@
...
+      {
+        return tl::make_unexpected("Conversion from FORMAT_RLE is not yet implemented.");
+      }
+      else
+      {
+        return tl::make_unexpected("Unknown image format.");
+      }
+}
+
+tl::expected<std::vector<geometry_msgs::msg::TransformStamped>, std::string> getImageTransforms(const bosdyn::api::ImageResponse& image_response, const std::string& robot_name, const google::protobuf::Duration& clock_skew)
```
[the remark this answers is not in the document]
jschornak-bdai: Fixed!

spot_driver_cpp/README.md
```diff
@@ -0,0 +1,3 @@
+# spot_driver_cpp
+
+ROS 2 nodes to interact with Spot through its C++ SDK.
```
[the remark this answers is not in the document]
jschornak-bdai: I added some more content to this README, let me know if you think it's sufficient.

spot_driver_cpp/launch/spot_image_publisher.launch.xml
jschornak-bdai: I ended up deleting this launch file.

jschornak-bdai:
@bhung-bdai and I did some testing together to make sure that the hand camera images do get publish, so I think there is not actually a bug there.

bhung-bdai (approved):
LGTM

skpawar1305:
hey, great work
can you also allow to use image transport, a flag is fine if you prefer the other for a composable nodes like
https://github.com/IntelRealSense/realsense-ros/blob/ros2-development/realsense2_camera/include/image_publisher.h

also, I had done the same thing (in a hacky way though)
https://github.com/RRL-ALeRT/spot_ros2/blob/nuc/spot_cpp_ros2/

jbarry-bdai:
Given how static transforms are handled by C++, I think that code is now correct.  Can you verify you've done the following tests?  If so then I'll approve!

jschornak-bdai:
> Given how static transforms are handled by C++, I think that code is now correct. Can you verify you've done the following tests? If so then I'll approve!

@jbarry-bdai yep, we did test those! @bhung-bdai witnessed my camera-in-hand testing as well as the use of the new CameraInfo messages to create point clouds. In addition to my manual testing, there are unit tests to exercise getting the authentication inputs from either parameters and environmetn variables.

jbarry-bdai (approved):

PR #149, opened 2023-10-02
[SW-526] Added an exception field to the ROS2 logger for SpotWrapper

bhung-bdai:
As referenced [here](https://github.com/boston-dynamics/spot-sdk/blob/6a03ae12056a74e65fd9715128e0c40762dfa7ba/python/bosdyn-client/src/bosdyn/client/async_tasks.py#L182), the logger on Spot expects a `logging.Logger` style logger with an exception field instead of the RCUtilsLogger. The RCUtilsLogger doens't have an `exception` method, so I just wrapped it with an exception method. 

This was HIL tested by calling the `_handle_error` in SpotWrapper's various classes that inherit from `AsyncPeriodicQuery`. This should log data for future failures in the robot's start up.

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -1,5 +1,6 @@
 ### Debug
 # from ros_helpers import *
+import logging
 import sys
 import threading
 import time
@@ -14,6 +15,9 @@ import builtin_interfaces.msg
 import rclpy
 import rclpy.time
 import tf2_ros
+from bdai_ros2_wrappers.logging import (
+    logs_to_ros,
+)
 from bdai_ros2_wrappers.single_goal_action_server import (
     SingleGoalActionServer,
 )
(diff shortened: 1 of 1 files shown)


amessing-bdai:
https://github.com/bdaiinstitute/ros_utilities/pull/23

bhung-bdai:
Changed this to use the new logging forwarding. At the moment, there will need to be a change to format the message in the exception field logging.Logger, but it's a small change and this won't be affected by changes to ros_utilities going onwards

mpickett-bdai (approved):
LGTM. I have to look at these changes to ros_utilities to get up to data, but this looks fine.

bhung-bdai:
double checked with the `spot_wrapper` logging. There isn't any other place where the `spot_wrapper` logger is doing anything that isn't built into `logging.Logger`. Going to merge this as the tests seem fine, but can easily roll this back if something changes, Will need to update the `gitman.yml` with the updated versions for `ros_utilities` (if not already updated) and `spot_ros2`.

PR #152, opened 2023-10-03
[N/A] Add dependabot config to update submodules

amessing-bdai:
This will configure dependabot to open PRs to update the submodules. It will check nightly. PRs will still have to be manually approved/merged.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+
+updates:
+  - package-ecosystem: "gitsubmodule"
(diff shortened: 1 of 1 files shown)


bhung-bdai (approved):
LGTM. Should double check with JW that it doesn't have an edge case, since he knows more about Dependabot, but it's not a big deal if it does have that edge case.

jiuguangw (approved):
Looks good to me - I don't have a lot of Dependbot experience but looks very reasonable to me.

PR #161, opened 2023-10-12
Standalone launch file for rviz

jbarry-bdai:
This creates a standalone launch file that can be used to launch rviz so it doesn't always have to happen in the driver.

It also removes point clouds and images from the default rviz that's launched - currently it's just the robot model and TF.  That can be changed if people preferred it a different way.

This is based off @jschornak-bdai 's pointcloud publishing branch to avoid conflicts (and also cuz that's when I wanted it).


spot_driver/launch/spot_driver.launch.py
```diff
@@ -201,17 +177,12 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
...
-        rviz_config_file = PathJoinSubstitution([FindPackageShare(THIS_PACKAGE), "rviz", "spot.rviz"]).perform(context)
-
-    rviz = launch_ros.actions.Node(
-        package="rviz2",
-        executable="rviz2",
-        name="rviz2",
-        arguments=["-d", rviz_config_file],
-        output="screen",
+    rviz = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource([FindPackageShare("spot_driver"), '/launch', '/rviz.launch.py']),
```
amessing-bdai, line 181:
```suggestion
        PythonLaunchDescriptionSource([FindPackageShare(THIS_PACKAGE), '/launch', '/rviz.launch.py']),
```

mpickett-bdai (dismissed):
I agree Andrew's suggestion, but besides that looks good to go.

jschornak-bdai:
We merged #147, so this can be updated onto the latest state of `main`.

jbarry-bdai:
Oh yeah I just need to update and test this.  I'll get to it this week unless someone else wants to do that sooner :-p

bhung-bdai:
> Oh yeah I just need to update and test this. I'll get to it this week unless someone else wants to do that sooner :-p



> Oh yeah I just need to update and test this. I'll get to it this week unless someone else wants to do that sooner :-p

I can probably take a crack at it tomorrow. Is there anything specific I need to do, other than use the new arguments/config parameters for `DepthRegistered`?

spot_driver/launch/rviz.launch.py
```diff
@@ -0,0 +1,82 @@
+# Copyright [2023] Boston Dynamics AI Institute, Inc.
```
bhung-bdai, line 1: The Quality of Service needs to be set to "best effort" in Rviz if you want to visualize the PointCloud2 under `Reliability Policy`. Is there a way to set that automatically in this config?

bhung-bdai (approved):
LGTM, tested on hardware and it's much easier to pull up than before. Wondering if we can set the QoS automagically, but if we can't we should update the README on how to visualize the point clouds somewhere.

spot_driver/launch/rviz.launch.py
jschornak-bdai: There is a way to have that default to the right setting, but I think you need to provide that in the config YAML file. Here's the full entry for a PointCloud display, for example. We set `Topic -> Reliability Policy` to `Best Effort` instead of `Reliable`.

```
    - Alpha: 1
      Autocompute Intensity Bounds: true
      Autocompute Value Bounds:
        Max Value: 3.0432052612304688
        Min Value: -0.07985210418701172
        Value: true
      Axis: Z
      Channel Name: intensity
      Class: rviz_default_plugins/PointCloud2
      Color: 255; 255; 255
      Color Transformer: AxisColor
      Decay Time: 0
      Enabled: true
      Invert Rainbow: false
      Max Color: 255; 255; 255
      Max Intensity: 4096
      Min Color: 0; 0; 0
      Min Intensity: 0
      Name: PointCloud2
      Position Transformer: XYZ
      Selectable: true
      Size (Pixels): 3
      Size (m): 0.009999999776482582
      Style: Flat Squares
      Topic:
        Depth: 5
        Durability Policy: Volatile
        History Policy: Keep Last
        Reliability Policy: Best Effort
        Value: /points_back
      Use Fixed Frame: true
      Use rainbow: true
      Value: true
```

bhung-bdai (approved):
HIL tested and passed

PR #164, opened 2023-10-19
[N/A] Convexified collision geometries

wrangel-bdai:
This improves the collision geometries of the Spot model by providing simplified convex meshes and primitive shapes for the feet.


bhung-bdai:
Getting these errors when I try to use Rviz with the new changes:
image
Looks like it's not supporting these `.obj` files

wrangel-bdai:
> LGTM, works fine with RViz

Great! Since I am not an authorized user, do you mind merging it?

amessing-bdai:
@wrangel-bdai Do we have an archived version of the originals? Just in case they are ever needed again?

RussTedrake:
> @wrangel-bdai Do we have an archived version of the originals? Just in case they are ever needed again?

Of course you can always get them from the git history, no?

bhung-bdai:
> > @wrangel-bdai Do we have an archived version of the originals? Just in case they are ever needed again?
> 
> Of course you can always get them from the git history, no?

I believe it works as long as you checkout the old version and build. That being said, it might be nice to include both of them just to be safest. If you need it quickly over the weekend @wrangel-bdai I can merge it now and we can adjust it on Monday; if not, I'd rather have both versions saved just in case.

RussTedrake:
I'm ok working from the PR. No need to rush on my account.

wrangel-bdai:
> > > @wrangel-bdai Do we have an archived version of the originals? Just in case they are ever needed again?
> > 
> > 
> > Of course you can always get them from the git history, no?
> 
> I believe it works as long as you checkout the old version and build. That being said, it might be nice to include both of them just to be safest. If you need it quickly over the weekend @wrangel-bdai I can merge it now and we can adjust it on Monday; if not, I'd rather have both versions saved just in case.

I just got back to my keys. The old collision geometries are easily recoverable, but also fundamentally wrong. Hence, its not worth to keep them around.

- Arm collision: Those were just the visual geometries exported as an stl file. (very easy to recover)
- Base collisions: Those were the visual geometries but buffered. So if you would simulated the robot standing on ground, the feet would be lifted by three centimeters.

I think we are good with merging this and abandoning what we had before :)

bhung-bdai:
> > > > @wrangel-bdai Do we have an archived version of the originals? Just in case they are ever needed again?
> > > 
> > > 
> > > Of course you can always get them from the git history, no?
> > 
> > 
> > I believe it works as long as you checkout the old version and build. That being said, it might be nice to include both of them just to be safest. If you need it quickly over the weekend @wrangel-bdai I can merge it now and we can adjust it on Monday; if not, I'd rather have both versions saved just in case.
> 
> I just got back to my keys. The old collision geometries are easily recoverable, but also fundamentally wrong. Hence, its not worth to keep them around.
> 
> * Arm collision: Those were just the visual geometries exported as an stl file. (very easy to recover)
> * Base collisions: Those were the visual geometries but buffered. So if you would simulated the robot standing on ground, the feet would be lifted by three centimeters.
> 
> I think we are good with merging this and abandoning what we had before :)

Sounds good. I'll put out a thing in the internal channel for anyone who's using it to let them know.

PR #169, opened 2023-10-23
Automatically detect if Spot has the arm installed.

gremigi-bdai:
With this PR, we do not need the argument `has_arm` anymore. When we execute `bdai spot driver <spot-name>` this parameter is retrieved form Spot itself using the `SpotWrapper` interface.


spot_driver/launch/spot_driver.launch.py
```diff
@@ -133,15 +131,25 @@ def create_rviz_config(robot_name: str) -> None:
 
 
 def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
+    logger = logging.getLogger("spot_driver_launch")
+
```
amessing-bdai, line 135: I think this was unintentionally left here

amessing-bdai (commented):
Overall looks good. A logger that isn't used was left.

spot_driver/launch/spot_driver.launch.py
gremigi-bdai: It is required few lines below by the `SpotWrapper`. I tried passing None but it fails at runtime.

jbarry-bdai (commented):
Yay!  Tests to do:

- [ ] Spot without arm launches
- [x] Spot without arm is shown without arm in rviz
- [x] Spot with arm launches
- [x] Spot with arm is shown with arm in rviz
- [x] For spot with arm all camera topics are published for the hand camera including depth registered (and we actually see images on these topics)

gremigi-bdai:
Spot with arm launches correctly and the arm is displayed in RVIZ.

image

bhung-bdai (approved):
LGTM. Reminder to make the changes to the cli args to deprecate the `has-arm` field when we run within the docker and to update the `gitman.yml` file in the main repo as well.

PR #173, opened 2023-10-25
[PN-68] Add ci job to run unit test

abaker-bdai:


.github/workflows/test.yml
```diff
@@ -31,6 +31,13 @@ jobs:
         source /opt/ros/$ROS_DISTRO/setup.bash
         colcon build --symlink-install
         source install/local_setup.bash
+    - name: Test Package
+      run: |
+        source /opt/ros/$ROS_DISTRO/setup.bash
+        source install/local_setup.bash
+        # skip testing submodules
+        colcon test --packages-skip bdai_ros2_wrappers --event-handlers console_cohesion+
```
jiuguangw, line 39: - Prefer `console_direct` vs `console_cohesion`
- Consider adding` --python-testing pytest --pytest-with-coverage` and add integration for Codecov
- Add a second line` colcon test-result --all --verbose`

jiuguangw (changes requested):

.flake8
```diff
@@ -0,0 +1,22 @@
+[flake8]
```
jiuguangw, line 1: I'm not sure why there's a mixture of flake8 and ruff. I'm out of the loop on the spot_ros2 project, but right now I see pre-commit, hooks, and their configurations in `pyproject.toml`. Is it possible to move these configurations into pyproject.toml?

Your concern about docstrings - our ruff config currently is only turned on for a select few checks. You could turn on the docstring check with:


```
[tool.ruff]
# Enable pycodestyle (`E`), Pyflakes (`F`), and import sorting (`I`)
select = ["E", "F", "I", "D"] # add "D" here
```

jiuguangw (changes requested):

.flake8
abaker-bdai: I saw that it was a part of the initial set of unit/lint tests for the `spot_driver` package but I can simply remove it since it technically isn't necessary and the pre-commit check should suffice.

As far as the docstrings, I plan on addressing that in another PR since I've seen that I need to manually change some of the ruff errors.

jiuguangw (approved):

PR #187, opened 2023-11-09
[PN-107] C++ - Refactor Spot Interfaces to enable easier expansion and improve testability

gremigi-bdai:


spot_driver_cpp/include/spot_driver_cpp/spot_image_publisher.hpp
```diff
@@ -94,10 +94,12 @@ class SpotImagePublisher {
...
   // Interface classes to interact with Spot and the middleware.
   std::unique_ptr<TimerInterfaceBase> timer_interface_;
-  std::unique_ptr<SpotInterfaceBase> spot_interface_;
+  std::unique_ptr<ImageApi> image_api_;
   std::unique_ptr<PublisherInterfaceBase> publisher_interface_;
   std::unique_ptr<ParameterInterfaceBase> parameter_interface_;
   std::unique_ptr<TfInterfaceBase> tf_interface_;
   std::unique_ptr<LoggerInterfaceBase> logger_interface_;
+
+  std::shared_ptr<Robot> robot_;
```
abaker-bdai, line 103: Should we consider passing ownership to the image_api? In conjunction with refactoring the use of `spot_image_sources` to the `image_api`? That way the publisher could be "Robot agnostic" and only need to interface with the `image_api`.

@jschornak-bdai thoughts?

gremigi-bdai: Because the Robot is used to retrieve the different clients, the main idea is to use it as a shared resourced that can be passed to different APIs, including the reverse kinematic API.

spot_driver_cpp/include/spot_driver_cpp/api/default_image_api.hpp
```diff
@@ -14,13 +16,10 @@ namespace spot_ros2 {
 /**
  * @brief Implements SpotInterfaceBase to use the Spot C++ SDK.
  */
-class SpotInterface : public SpotInterfaceBase {
+class DefaultImageApi : public ImageApi {
```
jschornak-bdai, line 19: I don't think this is an accurate use of the term "API" --  the library as a whole _provides_ an API. I think it's more correct to refer to the base class here as something like `ImageClientInterface`, since it's a pure virtual class, and then the child class can be `DefaultImageClient`.

spot_driver_cpp/include/spot_driver_cpp/spot_image_publisher.hpp
jschornak-bdai: We would only need to share the same instance of Robot between different interface implementations if we were creating a single executable which was an agglomeration of different Spot API clients. There's an important design question here -- will we be providing a single executable that exposes all the Spot API clients, or a collection of executables that each expose an individual client or closely-related set of clients?

My preference is to do the latter -- I do not think there is negative overhead in authenticating to Spot from multiple processes in parallel, and we can implement the nodes as component nodes in a container to reduce the ROS 2 middleware overhead which we'd otherwise incurred by having multiple DDS participants from multiple standalone nodes.

jschornak-bdai (changes requested):

spot_driver_cpp/include/spot_driver_cpp/images/images_middleware_handle.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <spot_driver_cpp/interfaces/rclcpp_parameter_interface.hpp>
+#include <spot_driver_cpp/interfaces/rclcpp_tf_interface.hpp>
+#include <spot_driver_cpp/interfaces/rclcpp_wall_timer_interface.hpp>
+#include <string>
+#include <tl_expected/expected.hpp>
+#include <unordered_map>
+
+namespace spot_ros2::images {
+
+struct ImagesMiddlewareHandle : public SpotImagePublisher::MiddlewareHandle {
```
jschornak-bdai, line 20: Since ImageMiddlewareHandle has private members, I think it's better to make it a `class` instead of a `struct`. This applies to the base class too.

```suggestion
class ImagesMiddlewareHandle : public SpotImagePublisher::MiddlewareHandle {
public:
```

spot_driver_cpp/test/src/images/test_spot_image_publisher_node.cpp
```diff
@@ -0,0 +1,136 @@
...
+  bool publish_depth_images = kDefaultPublishDepthImages;
+  bool publish_depth_registered_images = kDefaultPublishDepthRegisteredImages;
+  std::string spot_name;
+};
+
+class SpotImagePubNodeTestFixture : public ::testing::Test {
+ public:
+  void SetUp() override {
+    rclcpp::init(0, nullptr);
+    node = std::make_shared<rclcpp::Node>("test_image_publisher_node");
```
jschornak-bdai, line 63: The presence of these within the test is a sign that the implementation of these classes needs to be tweaked a little bit: we should be able to unit test SpotImagePublisherNode without having an actual instance of `rclcpp::Node` or initializing the ROS 2 context.

spot_driver_cpp/include/spot_driver_cpp/images/spot_image_publisher_node.hpp
```diff
@@ -4,16 +4,22 @@
...
-namespace spot_ros2 {
+namespace spot_ros2::images {
 /**
  * @brief Wraps SpotImagePublisher to allow using it like a rclcpp::Node.
  */
 class SpotImagePublisherNode {
  public:
+  SpotImagePublisherNode(std::shared_ptr<rclcpp::Node> node, std::unique_ptr<SpotApi> spot_api,
+                         const std::shared_ptr<ParameterInterfaceBase>& parameter_interface,
+                         const std::shared_ptr<LoggerInterfaceBase>& logger_interface);
```
jschornak-bdai, line 22: I think there needs to be a version of this constructor that takes a `SpotImagePublisher::MiddlewareHandle` as an argument to allow us to instantiate this class in tests without requiring a real instance of `rclcpp::Node`. This version of the constructor can create the ImagesMiddlewareHandle and then call the other constructor.

jschornak-bdai: It seems like there are only two cases we need to support here:
- We have a real Node, and we are creating rclcpp-specialized versions of the middleware handle and the other interface classes.
- We are in a unit test and we are using mock or fake versions of the middleware handle and interface classes.
So there isn't a situation where we need to mix having a real Node and fake/mock interface classes.

With that in mind, you can probably replace this constructor withe one like this:

```
SpotImagePublisherNode(std::unique_ptr<SpotImagePublisher::MiddlewareHandle> middleware_handle, std::unique_ptr<SpotApi> spot_api,
                         const std::shared_ptr<ParameterInterfaceBase>& parameter_interface,
                         const std::shared_ptr<LoggerInterfaceBase>& logger_interface)
```

spot_driver_cpp/src/images/spot_image_publisher_node.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+  internal_ = std::make_unique<SpotImagePublisher>(node_, spot_api_->image_client_api(), expected_has_arm.value());
+  internal_->initialize();
+}
+
+SpotImagePublisherNode::SpotImagePublisherNode(const rclcpp::NodeOptions& node_options) {
+  auto node = std::make_shared<rclcpp::Node>("image_publisher", node_options);
+  *this = SpotImagePublisherNode(node, std::make_unique<DefaultSpotApi>(kSDKClientName),
+                                 std::make_shared<RclcppParameterInterface>(node),
+                                 std::make_shared<RclcppLoggerInterface>(node->get_logger()));
```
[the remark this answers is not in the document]
gremigi-bdai: I think we should use a private initialization method which is used by both constructors. I read that *this can be dangerous when the object is already partially initialized.

jschornak-bdai (approved):

PR #196, opened 2023-11-16
Extension of Spot_Driver_cpp refactoring effort

abaker-bdai:
Branched this off of https://github.com/bdaiinstitute/spot_ros2/pull/187 to better track my changes to the current refactor design. 

I initially wanted to contain it to that PR but as I was updating the tests to reflect the PR's changes, I found more design changes that needed to be made so I wanted to capture them in their own PR.


spot_driver_cpp/include/spot_driver_cpp/api/spot_api.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <string>
+
+namespace spot_ros2 {
+
+class SpotApi {
+ public:
+  virtual tl::expected<void, std::string> createRobot(const std::string& ip_address, const std::string& robot_name) = 0;
+  virtual tl::expected<void, std::string> authenticate(const std::string& username, const std::string& password) = 0;
+  virtual tl::expected<bool, std::string> hasArm() const = 0;
+  virtual tl::expected<std::unique_ptr<ImageClientApi>, std::string> imageClient() const = 0;
```
kineticsystem, line 19: The problem I see with this is that we need to modify this class every time we add a new API. This interface could potentially become big.

spot_driver_cpp/src/api/default_image_client_api.cpp
```diff
@@ -224,9 +198,12 @@ tl::expected<std::vector<geometry_msgs::msg::TransformStamped>, std::string> get
 }  // namespace
 
 namespace spot_ros2 {
-DefaultImageApi::DefaultImageApi(std::shared_ptr<Robot> robot) : robot_{robot} {}
 
-tl::expected<GetImagesResult, std::string> DefaultImageApi::getImages(::bosdyn::api::GetImageRequest request) {
+DefaultImageClientApi::DefaultImageClientApi(::bosdyn::client::ImageClient* image_client,
+                                             std::shared_ptr<TimeSyncApi> time_sync_api, const std::string& robot_name)
+    : image_client_{std::move(image_client)}, time_sync_api_{time_sync_api}, robot_name_{robot_name} {}
```
kineticsystem, line 204: You are passing a raw pointer into a `unique_ptr`. This means that it is now your responsibility to destroy the client, is this intentional?  The use of `std::move` is not necessary for raw pointers. `std::move` means that you take the guts out of an object and throw them into another object, so the source object becomes unusable. But a raw pointer has no guts.

spot_driver_cpp/src/api/default_spot_api.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+tl::expected<std::unique_ptr<ImageClientApi>, std::string> DefaultSpotApi::imageClient() const {
+  const auto image_client_result = robot_->EnsureServiceClient<::bosdyn::client::ImageClient>(
+      ::bosdyn::client::ImageClient::GetDefaultServiceName());
+  if (!image_client_result.status) {
+    return tl::make_unexpected("Failed to create image client.");
+  }
+
+  return std::make_unique<DefaultImageClientApi>(image_client_result.response, time_sync_api_, robot_name_);
```
kineticsystem, line 71: This has to do with a previous comment. Currently, the `::bosdyn::client::ImageClient` is held by the robot, and it is shared to the external world using a raw pointer.

If we call this method twice, and we take ownership of the created object using a `unique_ptr` (which we do), this creates a potential memory leak. Destroying any of the two `unique_ptr`, will destroy the object. 

The `imageClient` method is not a factory, because part of the object is held by the robot.

I suggest the following:

- add a private `shared_ptr<ClientApi>` field to the class `DefaultSpotApi`.
- update the `image_client` method to return a  `shared_ptr<ClientApi>`.
- update the `image_client` method to create a `shared_ptr<ClientApi>` only if it has not been created before.
- with this change, the ClientApi is destroyed only once, so it is safe to convert the raw pointer into a unique_ptr<::bosdyn::client::ImageClient> and pass it to the constructor of the `DefaultClientApi`. This will also be destroyed once.

kineticsystem:
```cpp
tl::expected<std::shared_ptr<ImageClientApi>, std::string> DefaultSpotApi::imageClient() const {
  if (!image_client_) {
    const auto image_client_result = robot_->EnsureServiceClient<::bosdyn::client::ImageClient>(
        ::bosdyn::client::ImageClient::GetDefaultServiceName());
    if (!image_client_result.status) {
      return tl::make_unexpected("Failed to create image client.");
    }
    auto bdai_client = std::make_unique<::bosdyn::client::ImageClient>(image_client_result.response);
    image_client_ = std::make_shared<DefaultImageClientApi>(std::move(bdai_client), time_sync_api_, robot_name_);
  } else {
    return image_client_;
  }
}
```

spot_driver_cpp/include/spot_driver_cpp/api/spot_api.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <tl_expected/expected.hpp>
+
+#include <memory>
+#include <string>
+
+namespace spot_ros2 {
+
+class SpotApi {
+ public:
+  virtual tl::expected<void, std::string> createRobot(const std::string& ip_address, const std::string& robot_name) = 0;
```
kineticsystem, line 16: I suggest renaming this into 'initialize_robot', as it is not a factory anymore.

kineticsystem (approved):

PR #203, opened 2023-12-03
Convert visual meshes to obj

wrangel-bdai:
This improves the textures of the spot base and adds missing ones to the arm. For consistency sake, I also converted the accessories to obj files.


bhung-bdai:
As long as the RViz render worked, I don't see any reason not to approve it

bhung-bdai (approved):
RViz works fine.

PR #219, opened 2023-12-22
Add ROS service to take new Spot lease

ashay-bdai:
To support shared autonomy with multiple clients (e.g. `bosdyn.client.RobotCommandClient`), we need a way to ask `SpotROS` to take a new lease.

The underlying SpotWrapper exposes a `getLease` interface, which either takes or acquires the lease, depending on the configuration for the driver. This PR uses the underlying interface of SpotWrapper directly to expose a ROS service that takes the lease.

For example, a teleop agent may have taken control of Spot via `lease_client.take()`, and a skill commanding spot through `SpotROS` will need to ask `SpotROS` to generate a new lease when teleop is done in order to be able to control Spot again (because`SpotROS` would be running with `use_take_lease=True` and `get_lease_on_action=False`).

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -959,6 +959,27 @@ class SpotROS(Node):
                 if self.auto_stand.value:
                     self.spot_wrapper.stand()
 
+        self.create_service(
+            srv_type=Trigger,
+            srv_name="take_or_acquire_lease",
+            callback=self.take_or_acquire_lease_callback,
+            callback_group=self.group,
+        )
+
+    def take_or_acquire_lease_callback(self, request: Trigger.Request, response: Trigger.Response) -> Trigger.Response:
+        self.get_logger().info("Incoming request to take or acquire a new lease.")
+        if self.spot_wrapper is None:
+            response.success = True
+            response.message = "spot_ros2 is running in mock mode."
+            return response
+
+        old_lease = self.spot_wrapper._lease
+        self.spot_wrapper.getLease()
+        lease = self.spot_wrapper._lease
+        response.success = str(lease.lease_proto) != str(old_lease.lease_proto)
+        response.message = str(lease.lease_proto)
+        return response
+
     def robot_state_callback(self, results: Any) -> None:
         """Callback for when the Spot Wrapper gets new robot state data.
         Args:
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -975,16 +975,25 @@ class SpotROS(Node):
             return response
 
         old_lease = self.spot_wrapper.lease2
-        # take() can technically raise an exception (haven't observed it, very rare),
+        # take() can technically raise an exception (although the two possibilities
+        # in the documentation don't seem to apply when take() is not given an argument),
         # but handling exceptions inside a ROS callback is overcomplicated,
         # so we ignore this for now.
         lease = self.spot_wrapper._lease_client.take()
         self.spot_wrapper._lease_keepalive = LeaseKeepAlive(self.spot_wrapper._lease_client)
         # There is no evidence that take() can give the same lease as before,
-        # but because many aspects of the spot sdk have been surprising and
+        # but because aspects of the spot sdk have been surprising and
         # undocumented, we do this check to be extra safe.
-        response.success = True if old_lease is None else str(lease.lease_proto) != str(old_lease.lease_proto)
-        response.message = str(lease.lease_proto)
+        have_new_lease = (old_lease is None and lease is not None) or (
+            str(lease.lease_proto) != str(old_lease.lease_proto)
+        )
+        if have_new_lease:
+            response.success = True
+            response.message = str(lease.lease_proto)
+        else:
+            response.success = False
+            response.message = ""
+
         return response
 
     def robot_state_callback(self, results: Any) -> None:
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -37,8 +37,8 @@ from bosdyn.api.geometry_pb2 import Quaternion, SE2VelocityLimit
 from bosdyn.api.spot import robot_command_pb2 as spot_command_pb2
 from bosdyn.api.spot.choreography_sequence_pb2 import Animation, ChoreographySequence
 from bosdyn.client import math_helpers
-from bosdyn.client.lease import LeaseKeepAlive
 from bosdyn.client.exceptions import InternalServerError
+from bosdyn.client.lease import LeaseKeepAlive
 from bosdyn_msgs.msg import (
     ArmCommandFeedback,
     Camera,
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -37,6 +37,7 @@ from bosdyn.api.geometry_pb2 import Quaternion, SE2VelocityLimit
 from bosdyn.api.spot import robot_command_pb2 as spot_command_pb2
 from bosdyn.api.spot.choreography_sequence_pb2 import Animation, ChoreographySequence
 from bosdyn.client import math_helpers
+from bosdyn.client.lease import LeaseKeepAlive
 from bosdyn.client.exceptions import InternalServerError
 from bosdyn_msgs.msg import (
     ArmCommandFeedback,
@@ -961,22 +962,22 @@ class SpotROS(Node):
 
         self.create_service(
             srv_type=Trigger,
-            srv_name="take_or_acquire_lease",
-            callback=self.take_or_acquire_lease_callback,
+            srv_name="take_lease",
+            callback=self.take_lease_callback,
             callback_group=self.group,
         )
 
-    def take_or_acquire_lease_callback(self, request: Trigger.Request, response: Trigger.Response) -> Trigger.Response:
-        self.get_logger().info("Incoming request to take or acquire a new lease.")
+    def take_lease_callback(self, request: Trigger.Request, response: Trigger.Response) -> Trigger.Response:
+        self.get_logger().info("Incoming request to take a new lease.")
         if self.spot_wrapper is None:
             response.success = True
             response.message = "spot_ros2 is running in mock mode."
             return response
 
-        old_lease = self.spot_wrapper._lease
-        self.spot_wrapper.getLease()
-        lease = self.spot_wrapper._lease
-        response.success = str(lease.lease_proto) != str(old_lease.lease_proto)
+        old_lease = self.spot_wrapper.lease2
+        lease = self.spot_wrapper._lease_client.take()
+        self.spot_wrapper._lease_keepalive = LeaseKeepAlive(self.spot_wrapper._lease_client)
+        response.success = True if old_lease is None else str(lease.lease_proto) != str(old_lease.lease_proto)
         response.message = str(lease.lease_proto)
         return response
 
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -975,15 +975,14 @@ class SpotROS(Node):
             return response
 
         old_lease = self.spot_wrapper.lease2
-        # take() can technically raise an exception (although the two possibilities
+        # `take()` can technically raise an exception (although the two possibilities
         # in the documentation don't seem to apply when take() is not given an argument),
         # but handling exceptions inside a ROS callback is overcomplicated,
         # so we ignore this for now.
         lease = self.spot_wrapper._lease_client.take()
         self.spot_wrapper._lease_keepalive = LeaseKeepAlive(self.spot_wrapper._lease_client)
-        # There is no evidence that take() can give the same lease as before,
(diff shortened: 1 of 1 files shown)


ashay-bdai:
* `main`
  * **PR #219** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/219?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/219?utm_source=stack-comment).

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -844,6 +844,31 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
...
                     self.spot_wrapper.stand()
 
+        # Explicitly get a lease when get_lease_on_action is False
+        # because spot_wrapper does not get a lease at initialization.
+        if self.spot_wrapper is not None and not self.get_lease_on_action.value:
+            self.spot_wrapper.getLease()
+        self.create_service(
+            srv_type=Trigger,
+            srv_name="gen_new_spot_ros_lease",
+            callback=self.new_lease_callback,
```
amessing-bdai, line 854: Maybe:
```suggestion
            srv_name="get_lease",
            callback=self.get_lease_callback,
```

to keep things consistent and make it clear what the service is calling under the hood?

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -844,6 +844,31 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
...
+            callback_group=self.group,
+        )
+
+    def new_lease_callback(self, request: Trigger.Request, response: Trigger.Response) -> Trigger.Response:
+        self.get_logger().info("Incoming request to generate a new lease.")
+        if self.spot_wrapper is None:
+            response.success = True
+            response.message = "SpotROS is running in mock mode."
```
amessing-bdai, line 862: nit: Use either `spot_ros2` (the node name) or `spot_driver` (the package name)

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -844,6 +844,31 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
...
+            return response
+
+        old_lease = self.spot_wrapper._lease
+        self.spot_wrapper.getLease()
```
amessing-bdai, line 866: Should we update `getLease` to return the lease? It wouldn't break existing code and then you wouldn't need to access a private member here

amessing-bdai (commented):
Some small things

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -844,6 +844,31 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
                 if self.auto_stand.value:
                     self.spot_wrapper.stand()
 
+        # Explicitly get a lease when get_lease_on_action is False
+        # because spot_wrapper does not get a lease at initialization.
+        if self.spot_wrapper is not None and not self.get_lease_on_action.value:
+            self.spot_wrapper.getLease()
```
ggoretkin-bdai, line 850: What is the consequence of not adding this change? I think the spirit of this PR could be that if you don´t care about this new functionality (ROS service to generate new lease), then there is no functional change to `spot_ros2`.

spot_driver/spot_driver/spot_ros2.py
ggoretkin-bdai: I'm in favor of consistency. I'd also like to make it clear that this is not a "getter" in the sense of just returning some information, but also triggers what BD API calls either "take" or "acquire":

https://github.com/bdaiinstitute/spot_wrapper/blob/5bcf6d53abc6c9ae2fe8dd9ef8cadfeb1ca52e60/spot_wrapper/wrapper.py#L1070

question: do we want to expose to ROS these two different "get"s individually?

spot_driver/spot_driver/spot_ros2.py
ggoretkin-bdai: I think that could be a good idea, but also raises potential confusion that this is just a getter instead of an "action" in the BD API.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -844,6 +844,31 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
...
+        self.get_logger().info("Incoming request to generate a new lease.")
+        if self.spot_wrapper is None:
+            response.success = True
+            response.message = "SpotROS is running in mock mode."
+            return response
+
+        old_lease = self.spot_wrapper._lease
+        self.spot_wrapper.getLease()
+        lease = self.spot_wrapper._lease
+        response.success = str(lease.lease_proto) != str(old_lease.lease_proto)
```
ggoretkin-bdai, line 868: Does the BD API itself provide a notion of success/failure here instead of trying to define a new notion of success? How would `self.spot_wrapper._lease` not change after a call to `self.spot_wrapper.getLease()`? I think only if the BD python API throws one of the exceptions corresponding to the error responses:

https://dev.bostondynamics.com/protos/bosdyn/api/proto_reference.html#bosdyn-api-TakeLeaseResponse-Status

https://dev.bostondynamics.com/protos/bosdyn/api/proto_reference.html#bosdyn-api-AcquireLeaseResponse-Status

What happens when one of the exceptions is thrown?

spot_driver/spot_driver/spot_ros2.py
ashay-bdai: @ name of service
I think `get_lease` could be confusing, especially in the context of the semiautonomy setting where many different nodes are "taking" the lease (generating new leases) -- someone might think `get_lease` doesn't generate a new lease, when it actally does. 

What about `spot_wrapper_get_lease`, or just something that references spot_wrapper somehow to make it clearer that it's spot_wrapper's get_lease?

@ exposing two different "get"s 
Maybe we can wait until someone has a use case for getting the current lease `SpotROS` is using, without generating a new one? I don't see anyone needing this in a semiautonomy setting where only one ros node can command Spot at a time. Will there be a scenario where we want two ros nodes to be able to command Spot simultaneously, where one commands it via `SpotROS`, and another commands it via its own `RobotCommandClient`? In such a case, the latter node might want to use the same lease `SpotROS` is using, and so will want to ask for it via a "get". Perhaps such a scenario is if we want a navigation node to control Spot's legs, while some other node control's Spot's arm? Would we want to do anything like this any time soon? If not, we can not expose both "get"s.

spot_driver/spot_driver/spot_ros2.py
ashay-bdai: I support having `getLease` return the lease. I didn't do it originally because I thought I'd make minimal changes in the first go. 

I think the right way to resolve the confusion is to rename it, to something like `genLease`.

spot_driver/spot_driver/spot_ros2.py
ashay-bdai: I can remove this. Not having this just means that anyone commanding through `spot_ros2`, when running `spot_ros2` with `use_take_lease=True` and `get_lease_on_action=False`, will have to ask `spot_ros2` to explicitly create a new lease before it sends any commands to it. But that's fine.

spot_driver/spot_driver/spot_ros2.py
ashay-bdai: https://dev.bostondynamics.com/_modules/bosdyn/client/lease#LeaseResponseError

`LeaseClient.take()` can raise `InvalidResourceError`, or `NotAuthoritativeServiceError`. We don't encounter the former, because if we have no argument to `take()`, as in the call inside `SpotWrapper.getLease()`, it defaults to 'body' which I think is always a valid resource. 

And I have no idea when the latter comes up -- I don't understand it. 

So yeah, I can put this in a try/except if that is preferred. 

I think I was thinking that someone might change `SpotWrapper.getLease` at some point in a way that returns the current lease and not necessarily a new lease, and only checking for these exceptions would not catch that. I at least want to assert that the lease returned by `SpotWrapper.getLease()` is different than the lease before the call to that method.

spot_driver/spot_driver/spot_ros2.py
ashay-bdai: Sounds good, changed to `spot_ros2`.

spot_driver/spot_driver/spot_ros2.py
ggoretkin-bdai: I am in favor of reducing confusion, and I think the right way to do that is to avoid introducing terminology. The BD API terminology regarding this is "take" and "acquire", and wrapping those two terms as "get" or "generate" or any other name seems to be unnecessary confusion.

I imagine there is a convenience reason to have both "take" and "acquire" wrapped under one name, and the chosen option depending on `SpotWrapper` initialization:

```
            use_take_lease: Use take instead of acquire to get leases. This will forcefully take the lease from any other lease owner.
```.

Can we just call it "takeOrAcquireLease"?

Comment (amessing-bdai) [reply spot_driver/spot_driver/spot_ros2.py:866]: > I am in favor of reducing confusion, and I think the right way to do that is to avoid introducing terminology. The BD API terminology regarding this is "take" and "acquire", and wrapping those two terms as "get" or "generate" or any other name seems to be unnecessary confusion.
> 
> I imagine there is a convenience reason to have both "take" and "acquire" wrapped under one name, and the chosen option depending on `SpotWrapper` initialization:
> 
> ```
>             use_take_lease: Use take instead of acquire to get leases. This will forcefully take the lease from any other lease owner.
> ```
> 
> @amessing-bdai Can we just call it `takeOrAcquireLease`, and deprecate `getLease`?

We can update the function name, but would also need to check/PR  spot_ros (as in ros 1).

File: spot_driver/spot_driver/spot_ros2.py (hunk 1 above)
Comment (amessing-bdai) [reply spot_driver/spot_driver/spot_ros2.py:854]: Answered the first 2 parts below - I am ok with changing the function in spot_wrapper, so long as we make sure both spot_ros2 and spot_ros are fixed (will only need to check with the maintainer of spot_ros to see if there are any issues on their end).

As for the last paragraph, I don't think we need that until we have a use case. If you have 2 nodes that want to control a spot then they can both go through one `spot_driver`. Planning for an alternative interface from ros to the spot-sdk is outside the scope of this repo.

File: spot_driver/spot_driver/spot_ros2.py (hunk 3 above)
Comment (ggoretkin-bdai) [reply spot_driver/spot_driver/spot_ros2.py:866]: I put up https://github.com/bdaiinstitute/spot_wrapper/pull/83https://github.com/bdaiinstitute/spot_wrapper/pull/83 which exposes `_lease` as `lease2` (`lease` was taken. better name suggestions are welcome)

We can worry about updating the function name and deprecating the old uses separately.

**Comment by ggoretkin-bdai** [comment]
Thanks @ashay-bdai for renaming the service to match the BD API terminology of "take" and "acquire". I think this is good to go. @amessing-bdai @bhung-bdai ?

**Review by ggoretkin-bdai — APPROVED** [review]
Once https://github.com/bdaiinstitute/spot_wrapper/pull/83 is in, we can remove the `_lease` "private" field access, but I do not think that should block this PR.

**Review by amessing-bdai — APPROVED** [review]

File: spot_driver/spot_driver/spot_ros2.py [hunk 6]
```diff
@@ -845,6 +846,27 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
...
+        self.get_logger().info("Incoming request to take a new lease.")
+        if self.spot_wrapper is None:
+            response.success = True
+            response.message = "spot_ros2 is running in mock mode."
+            return response
+
+        old_lease = self.spot_wrapper.lease2
+        lease = self.spot_wrapper._lease_client.take()
+        self.spot_wrapper._lease_keepalive = LeaseKeepAlive(self.spot_wrapper._lease_client)
+        response.success = True if old_lease is None else str(lease.lease_proto) != str(old_lease.lease_proto)
```
ggoretkin-bdai, line 866:
```suggestion
        have_new_lease =  (old_lease is None and lease is not None) or (str(lease.lease_proto) != str(old_lease.lease_proto)
        if have_new_lease:
            response.success = True
            response.message = str(lease.lease_proto)
        else:
            response.success = False
            response.message =
```

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -845,6 +846,27 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
...
+        response.message = str(lease.lease_proto)
```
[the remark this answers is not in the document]
ashay-bdai: `lease` will never be `None` after calling `take()`, so we don't really need to check for that -- we'll always have a new lease. That is, if `lease` is None, then `take()` failed, and if it fails it raises an exception, which we are not handling, and the ros process will crash (it is complicated / no possible to handle exceptions in ros callbacks as far as I know). So, the suggested code won't behave differently than the current code. But, happy to change to this if preferred.

ggoretkin-bdai (approved):

PR #239, opened 2024-01-23
[WUD-1123] Add support for ptz wrapper

echun-bdai:
Add ros2 services for the PTZWrapper, MediaLogWrapper, and LightingWrapper in the Spot CAM wrapper of spot_wrapper.  This enables control of the ptz mount and enables high resolution image fetching with optional led illumination.

Proposed changes:
- Relevant services in spot_ros2
- .srv message definitions in spot_msgs

Testing done:
- [x] Test that each service produces expected behavior using a spot robot with Spot CAM payload.
- [x] Test that each service does not interfere with operation of a spot robot without Spot CAM payload.


bhung-bdai:
Have we looked at integrating the work for the `spot_ros2` testing machinery here? Would be nice to have some test cases

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -1442,6 +1556,272 @@ def handle_set_volume(self, request: SetVolume.Request, response: SetVolume.Resp
...
+        except Exception as e:
+            response.success = False
+            response.message = f"Error: {e}"
+            return response
+
+    def handle_get_led_brightness(
+        self, request: GetLEDBrightness.Request, response: GetLEDBrightness.Response
+    ) -> GetLEDBrightness.Response:
+        """Ros service handler for getting the current brightness of the Spot CAM onboard LEDs"""
+        if self.spot_cam_wrapper is None:
```
bhung-bdai, line 1792: I hate to nitpick here, but is there a way we can do this without copy pasting this block?

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -1442,6 +1556,272 @@ def handle_set_volume(self, request: SetVolume.Request, response: SetVolume.Resp
...
+            return response
+
+        try:
+            proto_position = self.spot_cam_wrapper.ptz.get_ptz_position(request.name)
+            conv.convert_proto_to_bosdyn_msgs_ptz_position(proto_position, response.position)
+            response.success = True
+            response.message = "Success"
+            return response
+        except Exception as e:
+            response.success = False
```
bhung-bdai, line 1598: Again, consider wrapping this in a function

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -1442,6 +1556,272 @@ def handle_set_volume(self, request: SetVolume.Request, response: SetVolume.Resp
...
+            return response
+
+        try:
+            proto_descriptions = self.spot_cam_wrapper.ptz.list_ptz()
+            descriptions = []
+            for proto_description in proto_descriptions:
+                ros_msg = PtzDescription()
+                conv.convert_proto_to_bosdyn_msgs_ptz_description(proto_description, ros_msg)
+                descriptions.append(ros_msg)
+            response.success = True
```
bhung-bdai, line 1573: Have we considered wrapping this up in a function as well? It seems that the `response.success = True` and `response.message = "Success"` is repeated a lot as well

bhung-bdai (commented):
I realize that the old design pattern was probably flawed with the copy and pasted versions of the functions response values and checks, but I think it would establish a good precedent to start wrapping these up in functions to avoid massive code bloat.

echun-bdai:
> Have we looked at integrating the work for the `spot_ros2` testing machinery here? Would be nice to have some test cases

Not currently, is there a way to use the testing machinery with the spotcam payload?

echun-bdai:
Regarding the repeated `response.message = "Success"` and `response.success=True`, I'm not sure how well wrapping them in a function would work since each response is a different type.  Perhaps we could save a `spot_cam_dne_msg = "..."` error message and set this in a one liner like:
```
response.success, response.message = False, self.spot_cam_dne_msg if self.spot_cam_wrapper is None
```
but I'm not sure if that is better than the current?  I think the same goes for the other repeats.

bhung-bdai:
Maybe passing in the message string as a parameter? It might be out of scope for this particular PR but it would make it nice.

echun-bdai:
I moved the check for spot_cam_wrapper inside the try/except handler so that we only set response.success = False and response.message = <e> once.  Its not as clean as a function but I think it should be more concise.  I also don't think the penalty of raising an exception here should be too much of a concern?  I verified that this produces the same behavior as the prior commit on a spot robot.  What do you think?

spot_driver/spot_driver/spot_ros2.py
amessing-bdai: We could make a decorator, but I think that is beyond the scope of this PR

amessing-bdai:
> > Have we looked at integrating the work for the `spot_ros2` testing machinery here? Would be nice to have some test cases
> 
> Not currently, is there a way to use the testing machinery with the spotcam payload?

@bhung-bdai We can make that a ticket for a future PR.

spot_msgs/srv/SetPtzPosition.srv
```diff
@@ -0,0 +1,8 @@
+# Angles are probably in degrees
```
amessing-bdai, line 1: Can we check this and have a more definitive comment

bhung-bdai:
> Fix the one comment about the angle and will approve

Given the circumstances, I will approve as well once that gets pushed. Will need to make note of the stylistic edits at some point.

amessing-bdai (approved):

PR #246, opened 2024-01-30
[SW-632] Check config file path exists in spot_ros2

khughes-bdai:
## Change Overview
The goal of this PR is to check if the `config_file` launch argument for `spot_driver.launch.py` has been set, and if so, checks if it is a valid path. If this file is not valid, an exception is thrown. This happens before any nodes get launched so that there is no chance of partially launched "zombie nodes". 
## Testing Done
I tested by running the spot driver with a physical Spot, passing in a nonexistent config file path. In this case the driver exits before any nodes get launched. If the config file was a real file, or if no config file was passed in (for example, if the user chooses to set parameters through environment variables instead) the behavior is unchanged. 
Example of the failure state:
```
$ ros2 launch spot_driver spot_driver.launch.py config_file:=notarealfile.yaml
[INFO] [launch]: All log files can be found below /home/khughes/.ros/log/2024-01-31-10-17-41-390265-khughes-LT-900652
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Configuration file 'notarealfile.yaml' does not exist!
$ 
```

### DIFF
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -119,7 +119,7 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
     # if config_file has been set (and is not the default empty string) and is also not a file, do not launch anything.
     config_file_path = config_file.perform(context)
     if (config_file_path != "") and (not os.path.isfile(config_file_path)):
-        raise Exception("Configuration file '{}' does not exist!".format(config_file_path))
+        raise FileNotFoundError("Configuration file '{}' does not exist!".format(config_file_path))
 
     if not mock_enable:
         # Get parameters from Spot.
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -116,12 +116,12 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
     publish_point_clouds_config = LaunchConfiguration("publish_point_clouds")
     mock_enable = IfCondition(LaunchConfiguration("mock_enable", default="False")).evaluate(context)
 
-    # parameters can be set either through the configuration file or by setting environment variables. 
-    # Want to throw an exception if setting params through config file and the config file does not exist. 
+    # parameters can be set either through the configuration file or by setting environment variables.
+    # Want to throw an exception if setting params through config file and the config file does not exist.
 
     # config file has been set if it is not equal to the default (empty string)
     # TODO would be nice to have some way to access the default value, or a better way of checking if it was set by user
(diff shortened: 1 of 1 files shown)


khughes-bdai:
* `main`
  * **PR #246** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/246?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/246?utm_source=stack-comment).

spot_driver/launch/spot_driver.launch.py
```diff
@@ -116,6 +116,11 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
     publish_point_clouds_config = LaunchConfiguration("publish_point_clouds")
     mock_enable = IfCondition(LaunchConfiguration("mock_enable", default="False")).evaluate(context)
 
+    # if config_file has been set (and is not the default empty string) and is also not a file, do not launch anything.
+    config_file_path = config_file.perform(context)
+    if (config_file_path != "") and (not os.path.isfile(config_file_path)):
```
khughes-bdai, line 121: I would like a better way of checking the default value of `config_file` launch argument to avoid hard-coding values, but unfortunately there is no method in the `LaunchConfiguration` class that does this.

ggoretkin-bdai (approved):
LGTM!

spot_driver/launch/spot_driver.launch.py
tcappellari-bdai: nit: do you know if checking `isfile` on an empty string error out or return false? if it returns false we could remove the first clause and just check if the file exists

khughes-bdai: isfile on an empty string does return false. I opted  not to do that because there is the option to pass in parameters through environment variables instead of a config file. So if the user sets environment variables and chooses to not pass in a config file, they would not be able to run the launchfile anymore. Open to suggestions on better ways to handle this though

spot_driver/launch/spot_driver.launch.py
```diff
@@ -116,6 +116,11 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
...
+        raise Exception("Configuration file '{}' does not exist!".format(config_file_path))
```
amessing-bdai, line 122:
```suggestion
        raise FileNotFoundError("Configuration file '{}' does not exist!".format(config_file_path))
```

baxelrod-bdai (approved):

PR #209, opened 2023-12-07
[PN-97] Implement Spot's Robot State Publisher in C++ 

abaker-bdai:
This PR adds a new C++-based Spot state publisher node, which replaces the previous functionality in the Spot Python driver.

The default `spot_driver.launch.py` launch file has been updated to use the C++ publisher

### How to Test

Launch the Spot driver through its launch file, and observe that the topics containing info about Spot's state are being published.

PR #252, opened 2024-02-01
[SW-642] Make it obvious when spot is running out of battery

tcappellari-bdai:
Created a new `spot_alerts` node to monitor stats and throw a pop-up alert warning the user. It currently only monitors the battery states and throws a warning up if the percentage is <= 10%.
Also added the `low_battery` parameter flag to prevent constant pop up spamming.

Tested on a spot with low battery.

### DIFF
--- a/spot_driver/spot_driver/ros_helpers.py
+++ b/spot_driver/spot_driver/ros_helpers.py
@@ -1,5 +1,6 @@
 import os
 import time
+from tkinter import messagebox
 from typing import Any, Callable, List, Optional, Tuple, Union
 
 import builtin_interfaces.msg
@@ -27,7 +28,6 @@ from rclpy.node import Node
 from rclpy.parameter import Parameter
 from sensor_msgs.msg import CameraInfo, CompressedImage, Image, JointState
 from tf2_msgs.msg import TFMessage
-from tkinter import messagebox
 
 from spot_driver.conversions import convert_proto_to_bosdyn_msgs_manipulator_state
 from spot_msgs.msg import (  # type: ignore
@@ -532,7 +532,9 @@ def get_tf_from_world_objects(
     return tf_msg
 
 
-def get_battery_states_from_state(state: robot_state_pb2.RobotState, spot_wrapper: SpotWrapper, node: Node) -> BatteryStateArray:
+def get_battery_states_from_state(
+    state: robot_state_pb2.RobotState, spot_wrapper: SpotWrapper, node: Node
+) -> BatteryStateArray:
     """Maps battery state data from robot state proto to ROS BatteryStateArray message
     Args:
         state: Robot State proto
@@ -559,11 +561,7 @@ def get_battery_states_from_state(state: robot_state_pb2.RobotState, spot_wrappe
         battery_states_array_msg.battery_states.append(battery_msg)
 
         if battery_msg.charge_percentage <= 10 and not node.get_parameter("low_battery").value:
-            low_battery_param = Parameter(
-                'low_battery',
-                Parameter.Type.BOOL,
-                True
-            )
+            low_battery_param = Parameter("low_battery", Parameter.Type.BOOL, True)
             node.set_parameters([low_battery_param])
             messagebox.showwarning(
                 title="Warning: Low Battery {}".format(battery_msg.identifier),
--- a/spot_driver/spot_driver/ros_helpers.py
+++ b/spot_driver/spot_driver/ros_helpers.py
@@ -24,6 +24,7 @@ from geometry_msgs.msg import (
 from google.protobuf.timestamp_pb2 import Timestamp
 from nav_msgs.msg import Odometry
 from rclpy.node import Node
+from rclpy.parameter import Parameter
 from sensor_msgs.msg import CameraInfo, CompressedImage, Image, JointState
 from tf2_msgs.msg import TFMessage
 from tkinter import messagebox
@@ -531,7 +532,7 @@ def get_tf_from_world_objects(
     return tf_msg
 
 
-def get_battery_states_from_state(state: robot_state_pb2.RobotState, spot_wrapper: SpotWrapper) -> BatteryStateArray:
+def get_battery_states_from_state(state: robot_state_pb2.RobotState, spot_wrapper: SpotWrapper, node: Node) -> BatteryStateArray:
     """Maps battery state data from robot state proto to ROS BatteryStateArray message
     Args:
         state: Robot State proto
@@ -557,12 +558,18 @@ def get_battery_states_from_state(state: robot_state_pb2.RobotState, spot_wrappe
         battery_msg.status = battery.status
         battery_states_array_msg.battery_states.append(battery_msg)
 
-        if battery_msg.charge_percentage <= 0.1:
+        if battery_msg.charge_percentage <= 10 and not node.get_parameter("low_battery").value:
+            low_battery_param = Parameter(
+                'low_battery',
+                Parameter.Type.BOOL,
+                True
+            )
+            node.set_parameters([low_battery_param])
             messagebox.showwarning(
                 title="Warning: Low Battery {}".format(battery_msg.identifier),
                 message=(
-                    "Battery is at {}. Approximately {} minutes remaining.\n Please charge your Spot soon."
-                ).format(battery_msg.charge_percentage * 100, battery_msg.estimated_runtime.sec / 60),
+                    "Battery is at {} %. Approximately {} minutes remaining.\n Please charge your Spot soon."
+                ).format(battery_msg.charge_percentage, round(battery_msg.estimated_runtime.sec / 60)),
             )
 
     return battery_states_array_msg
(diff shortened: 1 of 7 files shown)


tcappellari-bdai:
* `main`
  * **PR #252** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/252?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/252?utm_source=stack-comment).

spot_driver/spot_driver/ros_helpers.py
```diff
@@ -556,6 +560,16 @@ def get_battery_states_from_state(state: robot_state_pb2.RobotState, spot_wrappe
...
 
+        if battery_msg.charge_percentage <= 10 and not node.get_parameter("low_battery").value:
+            low_battery_param = Parameter("low_battery", Parameter.Type.BOOL, True)
+            node.set_parameters([low_battery_param])
+            messagebox.showwarning(
+                title="Warning: Low Battery {}".format(battery_msg.identifier),
+                message=(
+                    "Battery is at {} %. Approximately {} minutes remaining.\n Please charge your Spot soon."
+                ).format(battery_msg.charge_percentage, round(battery_msg.estimated_runtime.sec / 60)),
+            )
```
jschornak-bdai, line 571: This is separate from my main comment, but it would be better to put this user-facing message higher up in the code, for example at the same level where `get_battery_states_from_state` is called in the `spot_ros2` node.

The `get_battery_states_from_state` function is a pretty low-level conversion function that's just responsible for converting a protobuf message to a ROS message, so it shouldn't pop up a UI element.

jschornak-bdai (changes requested):
As a heads-up, #209 is going to move the code that's modified in this PR into a separate C++-based node, which will impact your implementation here.

A future-proof way to handle this warning would be to make a ROS node that's added to the Spot driver launch description which subscribes to the battery state ROS topic published by the driver and runs the check you added here in the subscriber callback.

spot_driver/spot_driver/ros_helpers.py
tcappellari-bdai: ahhhh i forgot about the driver and driver_cpp merge. I think you're right and I should make a new node

spot_driver/spot_driver/spot_alerts.py
```diff
@@ -0,0 +1,52 @@
...
+                self.set_parameters([low_battery_param])
+                messagebox.showwarning(
+                    title="Warning: Low Battery {}".format(battery_msg.identifier),
+                    message=(
+                        "Battery is at {} %. Approximately {} minutes remaining.\n Please charge your Spot soon."
+                    ).format(battery_msg.charge_percentage, round(battery_msg.estimated_runtime.sec / 60)),
+                )
+
+
+@ros_process.main(prebaked=False)
```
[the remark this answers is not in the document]
tcappellari-bdai: i copied the syntax from another node since I'm not sure I know what a pre-baked Node is

spot_driver/spot_driver/spot_alerts.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def battery_callback(self, battery_array_msg: BatteryStateArray) -> None:
+        """A callback function to check the battery percentage and send a pop-up alert when <= 10%
+        Args:
+            battery_array_msg: BatteryStateArray message received from the publisher containing BatteryState messages
+        Returns:
+            None
+        """
+        for battery_msg in battery_array_msg.battery_states:
+            if battery_msg.charge_percentage <= 10 and not self.get_parameter("low_battery").value:
```
khughes-bdai, line 35: nit: might be nice to have 10 as a ros parameter vs hardcoded in case anyone wants to customize?

khughes-bdai (approved):
LGTM

spot_driver/spot_driver/spot_alerts.py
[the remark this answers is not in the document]
jschornak-bdai: I'm curious about this too -- @tcappellari-bdai do you have some time to investigate what this option does?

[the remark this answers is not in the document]
tcappellari-bdai: ya i can poke at it some more later this week

spot_driver/spot_driver/spot_alerts.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__("spot_alerts", **kwargs)
+        # Subscribers #
+        self.battery_states = self.create_subscription(
+            BatteryStateArray, "status/battery_states", self.battery_callback, 1
+        )
+
+        # Parameters #
+        self.declare_parameter("low_battery", False)
```
jbarry-bdai, line 25: Why do this as a ROS parameter and not just a class variable?

tcappellari-bdai: I made it a parameter to make it more easily accessible by other nodes if necessary (maybe some processes should automatically stop if spot has low battery?)

jbarry-bdai: I guess I would assume they can also just subscribe to the topic?  Feels more "normal".  But maybe this is more the ROS way?

tcappellari-bdai: ya for a flag like this I would recommend a parameter since other nodes can have a callback that is only triggered when a specific parameter is changed. A subscriber would run the callback every time they get a new message from the battery_states topic and would also create some duplicate code from this alerts node.

jschornak-bdai: Parameters work somewhat differently in ROS 2 vs. ROS 1 -- there isn't a globally-accessible parameter server anymore, so it's not simple for one node to retrieve a parameter from another node. I think subscribing to the battery state topic is the "ROS-y" way to communicate this across multiple nodes. Since the purpose of this specific node is to show a pop-up to the user when the battery level gets low, maybe it's not necessary to have this parameter.

spot_driver/spot_driver/spot_alerts.py
jschornak-bdai: ROS parameters are useful for setting user-configurable thresholds at startup, so I think this would be a good thing to expose as a parameter with 10 as the default value. You wouldn't necessarily need to modify the launchfile to allow the user to set it at runtime as part of this PR, either.

spot_driver/spot_driver/spot_alerts.py
tcappellari-bdai: The parameters live under the domain of whatever node created it now (as opposed to the master node) but can still be accessed by other nodes. It mostly means you have to specify the node name as well as the parameter name instead of in ROS 1 when you just needed the parameter name

spot_driver/spot_driver/spot_alerts.py
tcappellari-bdai: Ok fair enough. I just couldnt really think of a scenario in which the user would need to change the low battery threshold while writing this lol

jschornak-bdai: I don't have an especially strong opinion about this either way. I think it's OK as you have it written right now.

spot_driver/spot_driver/spot_alerts.py
jschornak-bdai: Expanding on my "not so simple" comment, it's _possible_ to do this but it's also a big pain that requires a fair amount of finicky code (like the async parameter client in [this demo](https://github.com/ros2/demos/blob/master/demo_nodes_cpp/src/parameters/set_and_get_parameters_async.cpp), for example). I'd generally suggest a different approach for communicating this data between nodes.

jschornak-bdai (approved):
I tested this out by publishing a BatteryStateArray message like this:

```
ros2 topic pub --rate 1 /status/battery_states spot_msgs/msg/BatteryStateArray "{battery_states: [{charge_percentage: 8.0}]}"
```

, and the pop-up appears like I'd expect.

There are a couple things that can be addressed in follow-up work, such as handling Spot charging back above the low-battery threshold level while these nodes are still running.

tcappellari-bdai:
> I tested this out by running the spot_alerts node by itself and publishing a BatteryStateArray message like this:
> 
> ```
> ros2 topic pub --rate 1 /status/battery_states spot_msgs/msg/BatteryStateArray "{battery_states: [{charge_percentage: 8.0}]}"
> ```
> 
> , and the pop-up appears like I'd expect.
> 
> There are a couple things that can be addressed in follow-up work, such as handling Spot charging back above the low-battery threshold level while these nodes are still running.

I just added in an extra check to reset the low_battery flag in case the spot is charging

jbarry-bdai (approved):
Sorry just realized in all the questions I had about parameters I never clicked approve.  Thanks for doing this!!

PR #234, opened 2024-01-19
[EK-25] Use `bosdyn_msgs` bundles

mhidalgo-bdai:
Depends on https://github.com/bdaiinstitute/bosdyn_msgs/pull/14. This patch updates this repository to work with the binary bundles that https://github.com/bdaiinstitute/bosdyn_msgs will release from now on. Changes are minimal (but changes nonetheless).

Tested manually by running examples against an Spot robot. I'm open to ideas on how to further test this.

PR #256, opened 2024-02-02
Rework CI to ensure up-to-date containers prior to builds

mhidalgo-rai:
This patch merges lint, docker image build, and package build and test workflows into a single one, so as to ensure that package builds are performed in up-to-date containers. Similar to https://github.com/bdaiinstitute/ros_utilities/pull/47. Necessary for #234.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          path: src/spot_ros2
+          path: src
       - name: Build packages
         run: |
           source /opt/ros/$ROS_DISTRO/setup.bash
@@ -95,7 +95,6 @@ jobs:
           colcon test --python-testing pytest --pytest-with-coverage --event-handlers console_direct+ --packages-skip-regex bdai_ros2_wrappers proto2ros
       - name: Generate coverage report
         run: lcov -c -d build/spot_driver/ -o coverage_spot_driver.info --include "*/spot_driver/*" --exclude "*/test/*"
-        working-directory: ${{ github.workspace }}/../../
       - name: Upload python coverage to Coveralls
         uses: coverallsapp/github-action@v2
         with:
@@ -104,7 +103,7 @@ jobs:
           flag-name: unittests-python
           parallel: true
           debug: true
-          files: $(find ../../ -name "coverage.xml" -type f)
+          files: $(find . -name "coverage.xml" -type f)
       - name: Upload cpp coverage to Coveralls
         uses: coverallsapp/github-action@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+  REGISTRY: ghcr.io
+  # github.repository as <account>/<repo>
+  IMAGE_NAME: bdaiinstitute/spot_ros2_jammy_humble
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  lint:
```
mhidalgo-bdai, line 27: Making the lint job a prerequisite for the rest ensures no cycles are spent testing a PR that will invariably need an update.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Build and push Docker image (may be cached)
+        uses: docker/build-push-action@ac9327eae2b366085ac7f6a2d02df8aa8ead720a  # https://github.com/docker/build-push-action
+        with:
+          context: .
+          file: .devcontainer/Dockerfile
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
mhidalgo-bdai, line 75: Layer caching will keep the cost of image re-builds at bay.

mhidalgo-bdai:
Dependencies still wrong. I'll rework that and push once https://github.com/bdaiinstitute/spot_wrapper/pull/85 lands.

mhidalgo-bdai:
Coveralls complains in `spot_wrapper` sources. I don't think that is related to this patch.

.devcontainer/Dockerfile
```diff
@@ -36,15 +29,6 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
 # I added this
 RUN source "/opt/ros/${ROS_DISTRO}/setup.bash"
 
-# Spot packages
```
bhung-bdai, line 39: Confused by these removals. This docker container is not just for CI; it's also provided to the end user in case they don't want to develop on their machine. Does removing these still let them run the ros packages? Is it getting installed during the ROSdep call downstream?

mhidalgo-bdai: > Does removing these still let them run the ros packages? Is it getting installed during the ROSdep call downstream?

That's correct. Packages in this repository already declare these as dependencies, with the exception of `vision_msgs`. That one is not referenced nor used by any IIRC.

bhung-bdai (approved):
This appears all well to me. Can we separate out the `spot_wrapper` tests for now, since they keep screwing the CI over?

mhidalgo-bdai:
>  Can we separate out the spot_wrapper tests for now, since they keep screwing the CI over?

@bhung-bdai do you want that to happen in this PR or in a follow-up one?

amessing-bdai (approved):

Issue #247, opened 2024-01-30
Config. file is not used to get hostname, username, or password when running the driver.


jacobbreen25:
When I try to run the spot_driver, despite putting a config file (which is below), I am still required to set the environmental variables to access the robot. Here is the console:

```
ros2 launch spot_driver spot_driver.launch.py spot_name:="spot" config_file:="./config.yaml" depth_registered_mode:=from_spot publish_point_clouds:=true launch_rviz:=true
[INFO] [launch]: All log files can be found below /home/jbreen/.ros/log/2024-01-30-15-27-05-663731-csrobot-B450M-DS3H-WIFI-8776
[INFO] [launch]: Default logging verbosity is set to INFO
Failed to communicate with robot: UnableToConnectToRobotError: The robot may be offline or otherwise unreachable.
Ensure the robot is powered on and you can ping . Robot may still be booting. Will retry in 15 seconds
Failed to communicate with robot: UnableToConnectToRobotError: The robot may be offline or otherwise unreachable.
Ensure the robot is powered on and you can ping . Robot may still be booting. Will retry in 15 seconds
^C[WARNING] [launch]: user interrupted with ctrl-c (SIGINT)
```
Here is the YAML contents:
```
/**:
  ros__parameters:
    username: "user"
    password: "REDACTED"
    hostname: "192.168.80.3"
    robot_state_rate: 10.0
    metrics_rate: 0.04
    lease_rate: 1.0
    image_rate: 10.0
    auto_claim: False
    auto_power_on: False
    auto_stand: False
    deadzone: 0.05
    estop_timeout: 9.0
    start_estop: False
    preferred_odom_frame: "odom"
    cmd_duration: 0.125
    rgb_cameras: True
    initialize_spot_cam: True
```
I moved the username, password, and hostname up to the top however, this was a copy of the example code and it still didn't work

orrblue:
@jacobbreen25 Seems like we're at about the same level of familiarity with this repo. What command did you use to run spot_driver when you encountered this issue?

jacobbreen25:
@orrblue I didn't realize it didn't copy the command above. Below is my command:

```
ros2 launch spot_driver spot_driver.launch.py spot_name:="spot" config_file:=./config.yaml depth_registered_mode:="from_spot" publish_point_clouds:=true launch_rviz:=true
```

bhung-bdai:
@khughes-bdai is working on that

khughes-bdai:
[...]

PR #259, opened 2024-02-05
[SW-679] Config. file is not used to get hostname, username, or password when running the driver

khughes-bdai:
## Changes Overview
This addresses [this issue](https://github.com/bdaiinstitute/spot_ros2/issues/247) of not being able to connect to Spot from `spot_driver.launch.py` without setting environment variables. Now, the spot driver can be launched with only a configuration file yaml (`config_file`) if the user chooses. 

The implementation parses the `config_file` yaml within the launchfile to determine if `username`,  `password`, and `hostname` are set there as ROS parameters. If both environment variables and `config_file` parameters are set, then the environment variables take priority (similar to the `get_from_env_and_fall_back_to_param` logic in `spot_ros2.py`).  It also assumes a yaml structure for `config_file` (if used) of the following:
```
/**:
    ros__parameters:
        username: ...
        password: ...
        hostname: xxx.xxx.xxx.xxx
        .....
```
and also assumes environment variables (if used) are set as `BOSDYN_CLIENT_USERNAME`, `BOSDYN_CLIENT_PASSWORD,` and `SPOT_IP`. 

## Testing Done
* If an environment variable (any of `BOSDYN_CLIENT_USERNAME`, `BOSDYN_CLIENT_PASSWORD,` `SPOT_IP`) is set, this will be used for the login attempt. 
* If any environment variable is not set, the `config_file` yaml will be searched to see if the appropriate login information exists there. 
* If a parameter of username/password/hostname is set both by environment variable and by `config_file`, the value from the environment variable will be used for the login attempt. 
* If any one of username, password, and hostname is not set by anything, an exception will be thrown as login will not be possible and the driver will not launch. 
* If the `config_file` yaml passed in is malformed and cannot be parsed, a YAMLError is thrown. 

Verified that login is possible with a config file on-robot too. 

## Other comments
Currently the only use for these variables in the launchfile is to instantiate SpotWrapper in order to test if the robot has an arm. This affects which nodes get launched (specifically the robot description and the camera/point cloud nodes to stream from the hand camera). I feel that this implementation results in a messy launchfile, and also means the robot is logged into twice (once in the launchfile, and then again in the main `spot_ros2` node). Parsing a configuration file by hand is the only way to get the value of these parameters in the launchfile, as the standard ROS way of getting parameters from a yaml is only available for nodes. It would be nice in the future to try to rework this to make for a more declarative launchfile, potentially using things like ActionHandlers.

### DIFF
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -130,7 +130,7 @@ def get_login_parameters(context: LaunchContext) -> Tuple[str, str, str, int]:
                         hostname = ros_params["hostname"]
             except yaml.YAMLError as exc:
                 print("Parsing config_file yaml failed with: {}".format(exc))
-    if (username is None) or (password is None) or (hostname is None):
+    if (not username) or (not password) or (not hostname):
         raise ValueError(
             "One or more of your login credentials has not been specified! Got invalid values of "
             "[Username: '{}' Password: '{}' Hostname: '{}']. Ensure that your environment variables are set or "
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -106,7 +106,7 @@ def create_point_cloud_nodelets(
 
 
 def spot_has_arm(context: LaunchContext) -> bool:
-    # Check if spot has an arm by logging in and instantiating a SpotWrapper
+    """Check if spot has an arm by logging in and instantiating a SpotWrapper"""
     config_file_path = LaunchConfiguration("config_file").perform(context)
     spot_name = LaunchConfiguration("spot_name").perform(context)
     logger = logging.getLogger("spot_driver_launch")
@@ -122,11 +122,11 @@ def spot_has_arm(context: LaunchContext) -> bool:
                 if ("/**" in config_dict) and ("ros__parameters" in config_dict["/**"]):
                     ros_params = config_dict["/**"]["ros__parameters"]
                     # only set username/password/hostname if they were not already set as environment variables.
-                    if (username is None) and ("username" in ros_params):
+                    if (not username) and ("username" in ros_params):
                         username = ros_params["username"]
-                    if (password is None) and ("password" in ros_params):
+                    if (not password) and ("password" in ros_params):
                         password = ros_params["password"]
-                    if (hostname is None) and ("hostname" in ros_params):
+                    if (not hostname) and ("hostname" in ros_params):
                         hostname = ros_params["hostname"]
             except yaml.YAMLError as exc:
                 print("Parsing config_file yaml failed with: {}".format(exc))
@@ -156,11 +156,11 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
     if (config_file_path != "") and (not os.path.isfile(config_file_path)):
         raise FileNotFoundError("Configuration file '{}' does not exist!".format(config_file_path))
 
-    if not mock_enable:
-        has_arm = spot_has_arm(context)
-    else:
+    if mock_enable:
         mock_has_arm = IfCondition(LaunchConfiguration("mock_has_arm")).evaluate(context)
         has_arm = mock_has_arm
+    else:
+        has_arm = spot_has_arm(context)
 
     pkg_share = FindPackageShare("spot_description").find("spot_description")
 
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2024 Boston Dynamics AI Institute LLC. All rights reserved.
+# Copyright (c) 2023-2024 Boston Dynamics AI Institute LLC. All rights reserved.
 
 import logging
 import os
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -1,4 +1,4 @@
-# Copyright [2023] Boston Dynamics AI Institute, Inc.
+# Copyright (c) 2024 Boston Dynamics AI Institute LLC. All rights reserved.
 
 import logging
 import os
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -7,6 +7,7 @@ from typing import List
 
 import launch
 import launch_ros
+import yaml
 from launch import LaunchContext, LaunchDescription
 from launch.actions import DeclareLaunchArgument, IncludeLaunchDescription, OpaqueFunction
 from launch.conditions import IfCondition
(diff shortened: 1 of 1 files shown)


khughes-bdai:
* `main`
  * **PR #259** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/259?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/259?utm_source=stack-comment).

bhung-bdai (approved):
Ah, I see the original launch didn't actually get from the configuration file. Good catch!

spot_driver/launch/spot_driver.launch.py
```diff
@@ -122,18 +157,7 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
         raise FileNotFoundError("Configuration file '{}' does not exist!".format(config_file_path))
 
     if not mock_enable:
```
baxelrod-bdai, line 159: nit: "notting" this condition is kind of pointless here.  I'd rather see a more straightforward `if mock_enable:`

spot_driver/launch/spot_driver.launch.py
```diff
@@ -104,9 +105,43 @@ def create_point_cloud_nodelets(
...
+    port = int(os.getenv("SPOT_PORT", "0"))  # TODO should the user be able to specify a port via config file?
+    # parse the yaml to determine if login information is set there
+    if os.path.isfile(config_file_path):
+        with open(config_file_path, "r") as config_yaml:
+            try:
+                config_dict = yaml.safe_load(config_yaml)
+                if ("/**" in config_dict) and ("ros__parameters" in config_dict["/**"]):
+                    ros_params = config_dict["/**"]["ros__parameters"]
+                    # only set username/password/hostname if they were not already set as environment variables.
+                    if (username is None) and ("username" in ros_params):
```
baxelrod-bdai, line 125: nit: It's a long shot, but maybe use `if (not username) and (...):` to protect against the case when it is an empty string?

spot_driver/launch/spot_driver.launch.py
```diff
@@ -104,9 +105,43 @@ def create_point_cloud_nodelets(
...
+            "Login to Spot failed using [Username: '{}' Password: '{}' Hostname: '{}']. Update your config_file yaml"
+            " or ensure that your environment variables are set.".format(username, password, hostname)
+        )
+    spot_wrapper = SpotWrapper(
+        username=username, password=password, hostname=hostname, port=port, robot_name=spot_name, logger=logger
+    )
+    return spot_wrapper.has_arm()
 
+
+def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
```
baxelrod-bdai, line 144: nit: change `ld` to `description`

spot_driver/launch/spot_driver.launch.py
```diff
@@ -104,9 +105,43 @@ def create_point_cloud_nodelets(
     return composable_node_descriptions
 
 
-def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
+def spot_has_arm(context: LaunchContext) -> bool:
+    # Check if spot has an arm by logging in and instantiating a SpotWrapper
```
baxelrod-bdai, line 109: nit: use proper docstring format

baxelrod-bdai (approved):
LGTM, some comments from the peanut gallery

spot_driver/launch/spot_driver.launch.py
khughes-bdai: This was part of the original implementation (not sure why it's saying I added this code), and using ld for LaunchDescription seems to be the standard for other launchfiles in this repo

jschornak-bdai:
> Possible alternatives to consider for the future are a) add a boolean has_arm launch argument, but this would change the user interface

It used to work like this, but we specifically removed the `has_arm` argument in #169. The motivation for that is the driver has enough info available internally to figure out if Spot has an arm or not, and it's much simpler if the user doesn't need to keep track of that on their end.

spot_driver/launch/spot_driver.launch.py
```diff
@@ -104,9 +105,43 @@ def create_point_cloud_nodelets(
...
+                        password = ros_params["password"]
+                    if (not hostname) and ("hostname" in ros_params):
+                        hostname = ros_params["hostname"]
+            except yaml.YAMLError as exc:
+                print("Parsing config_file yaml failed with: {}".format(exc))
+    if (username is None) or (password is None) or (hostname is None):
+        raise ValueError(
+            "Login to Spot failed using [Username: '{}' Password: '{}' Hostname: '{}']. Update your config_file yaml"
+            " or ensure that your environment variables are set.".format(username, password, hostname)
+        )
```
jschornak-bdai, line 137: I think it would be nice to have this logic to load and parse the config file as a few separate function. It'll make `spot_has_arm` easier to read.

Maybe something like this (warning: pseudocode!):

```
def spot_has_arm:
  username, password, hostname, port = get_authentication_parameters()  # try parsing env vars and config file values in here
  spot_wrapper = SpotWrapper(
        username=username, password=password, hostname=hostname, port=port, robot_name=spot_name, logger=logger
  )
  return spot_wrapper.has_arm()
```

jschornak-bdai (commented):
This is a very philosophical comment, but I think it would be a good future effort to design the Spot drive launch files to be purely declarative. As currently implemented these launch files are actually executing Spot driver code (by creating a SpotWrapper, for example), and while that gets the job done it's sort of intermixing two different categories of things that are intended to be kept separate.

It would probably be possible to achieve this using [launch event handlers](https://docs.ros.org/en/rolling/Tutorials/Intermediate/Launch/Using-Event-Handlers.html). If we found a way to do that these launch files could be written as XML, which would greatly simplify things here. That's a topic for another day, though.

khughes-bdai:
@jschornak-bdai Definitely agree with wanting to make these launchfiles more declarative, this is something I'm potentially interested in working on in the future! I think it would also make things a lot more readable and avoid multiple instantiations of SpotWrapper. Not having to parse yamls inside of launchfiles would also be a nice bonus :)

spot_driver/launch/spot_driver.launch.py
```diff
@@ -104,9 +105,52 @@ def create_point_cloud_nodelets(
...
+                    # only set username/password/hostname if they were not already set as environment variables.
+                    if (not username) and ("username" in ros_params):
+                        username = ros_params["username"]
+                    if (not password) and ("password" in ros_params):
+                        password = ros_params["password"]
+                    if (not hostname) and ("hostname" in ros_params):
+                        hostname = ros_params["hostname"]
+            except yaml.YAMLError as exc:
+                print("Parsing config_file yaml failed with: {}".format(exc))
+    if (username is None) or (password is None) or (hostname is None):
```
baxelrod-bdai, line 133: nit: maybe use the same `if (not username)` pattern here

spot_driver/launch/spot_driver.launch.py
```diff
@@ -3,10 +3,11 @@
 import logging
```
baxelrod-bdai, line 3: nit: we should update the copyright to the proper format: https://www.notion.so/theaiinstitute/Copyrights-a3c268b0583d4518826b9de690602d6c

baxelrod-bdai (approved):

baxelrod-bdai (approved):

PR #268, opened 2024-02-13
Add a comment to explain why we need the power command in conftest.py

jbarry-bdai:
## Change Overview

I was trying to reproduce the testing here but left out the `PowerCommand` because it seemed unnecessary.  Took me a while to realize why it was so important so I added a comment about it :)

## Testing Done

None


bhung-bdai (approved):

khughes-bdai (approved):

mhidalgo-bdai:
BTW if we are going to need a mock of the `PowerCommand` in most if not all tests, we may as well add it to the collection of built-in mocks (I.e. the ones that `MockSpot` brings in by default) over at https://github.com/bdaiinstitute/spot_wrapper.

PR #276, opened 2024-02-21
Move functions to convert between ROS and proto messages into top-level spot_ros2 namespace

jschornak-bdai:
## Change Overview

This changes the namespace of the `convertToRos` and `convertToProto` utility functions. Previously they were in specific namespaces depending on the type of messages they converted (e.g., `spot_ros2::kinematic_conversions::convertToRos`). A problem with this is that it requires the user to know all the namespaces in order to find a specific conversion function, which can result in developers inadvertently adding duplicate functions. After this change, they're all in the top-level `spot_ros2` namespace.

This also changes one input parameter to be passed by const reference instead of by const value. The parameter is `double` so there's no performance gain for avoiding the copy but it makes a warning in my IDE go away.

This also adds a parameterized test case for a message conversion that had lots of internal conditionals. This is mostly to get the coverage metric up a bit.

## Testing Done

It is sufficient to make sure these packages compile and CI passes. This doesn't change robot behavior.

### DIFF
--- a/spot_driver/test/src/conversions/test_common_conversions.cpp
+++ b/spot_driver/test/src/conversions/test_common_conversions.cpp
@@ -1,9 +1,19 @@
-// Copyright (c) 2023 Boston Dynamics AI Institute LLC. All rights reserved.
+// Copyright (c) 2023-2024 Boston Dynamics AI Institute LLC. All rights reserved.
 
+#include <gmock/gmock-matchers.h>
+#include <gmock/gmock-more-matchers.h>
+#include <gmock/gmock.h>
+#include <gtest/gtest-param-test.h>
 #include <gtest/gtest.h>
+#include <bosdyn_msgs/msg/arm_joint_position.hpp>
+#include <bosdyn_msgs/msg/inverse_kinematics_request.hpp>
 #include <spot_driver/conversions/common_conversions.hpp>
 
 #include <rclcpp/rclcpp.hpp>
+
+namespace {
+using ArmJointPosition = bosdyn_msgs::msg::ArmJointPosition;
+}
 
 namespace spot_ros2::test {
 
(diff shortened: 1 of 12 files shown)


spot_driver/test/src/conversions/test_common_conversions.cpp
```diff
@@ -78,46 +88,144 @@ TEST(TestCommonConversions, convertBosdynMsgsRequestHeaderToProto) {
...
-  ASSERT_EQ(rosMsg.el1_is_set, protoMsg.has_el1());
-  ASSERT_EQ(rosMsg.el1, protoMsg.el1().value());
-  ASSERT_EQ(rosMsg.wr0_is_set, protoMsg.has_wr0());
-  ASSERT_EQ(rosMsg.wr0, protoMsg.wr0().value());
-  ASSERT_EQ(rosMsg.wr1_is_set, protoMsg.has_wr1());
-  ASSERT_EQ(rosMsg.wr1, protoMsg.wr1().value());
+class ConvertBosdynMsgsArmJointPositionToProtoParameterized : public ::testing::TestWithParam<ArmJointPosition> {};
+
+TEST_P(ConvertBosdynMsgsArmJointPositionToProtoParameterized, CheckFieldsNotSet) {
+  ArmJointPosition ros_msg = GetParam();
```
gbrooks-bdai, line 101:
```suggestion
  // GIVEN an arm joint position message with some number of fields set or not set
  ArmJointPosition ros_msg = GetParam();
```

spot_driver/test/src/conversions/test_common_conversions.cpp
```diff
@@ -78,46 +88,144 @@ TEST(TestCommonConversions, convertBosdynMsgsRequestHeaderToProto) {
...
+  bosdyn::api::ArmJointPosition proto_msg;
+  convertToProto(ros_msg, proto_msg);
```
gbrooks-bdai, line 103:
```suggestion
  // WHEN converting to a protobuf message
  convertToProto(ros_msg, proto_msg);
```

spot_driver/test/src/conversions/test_common_conversions.cpp
```diff
@@ -78,46 +88,144 @@ TEST(TestCommonConversions, convertBosdynMsgsRequestHeaderToProto) {
...
+
+  EXPECT_THAT(proto_msg.has_sh0(), testing::Eq(ros_msg.sh0_is_set));
```
gbrooks-bdai, line 105:
```suggestion
  // THEN we expect that each corresponding field is set an has the same value
  EXPECT_THAT(proto_msg.has_sh0(), testing::Eq(ros_msg.sh0_is_set));
```

gbrooks-bdai (changes requested):

gbrooks-bdai (approved):

PR #288, opened 2024-02-29
[SW-704] Remove special case for arm0.wr1 in spot_ros2

tcappellari-bdai:
## Change Overview

Removes the special handling for `arm0.wr1` in `spot_ros2` since the new SDK 4.0.0 fixed the issue.

## Testing Done

Spot driver and examples still run as expected


tcappellari-bdai:
* **#288** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/288?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/288?utm_source=stack-comment).

khughes-bdai (approved):
LGTM

jbarry-bdai (changes requested):
I should have put this in the issue but here's the specific test to run to make sure that they really did fix this issue:
1. Pull up rviz and visualize the point cloud from spot's hand in the vision frame
2. Move the hand around and ensure that the point cloud does not move around in the world and that link_wr1 tracks the hand correctly
3. Print the list of static transforms this is creating and ensure that there is a transform from link_wr1 to the camera frames and _not_ a link from body to link_wr1.  If there is that link then we need to add it to excluded frames.

tcappellari-bdai:
> I should have put this in the issue but here's the specific test to run to make sure that they really did fix this issue:
> 
> 1. Pull up rviz and visualize the point cloud from spot's hand in the vision frame
> 2. Move the hand around and ensure that the point cloud does not move around in the world and that link_wr1 tracks the hand correctly
> 3. Print the list of static transforms this is creating and ensure that there is a transform from link_wr1 to the camera frames and _not_ a link from body to link_wr1.  If there is that link then we need to add link_wr1 to excluded frames.

@jbarry-bdai the point cloud looks good on rviz (doesn't move with the arm), `arm0.link_wr1` appears to track the hand correctly, and I'm able to get the transforms between `arm0.link_wr1` and the other cameras; however `arm0.link_wr1` still has `body` as its parent frame.

tcappellari-bdai:
> @jbarry-bdai the point cloud looks good on rviz (doesn't move with the arm), `arm0.link_wr1` appears to track the hand correctly, and I'm able to get the transforms between `arm0.link_wr1` and the other cameras; however `arm0.link_wr1` still has `body` as its parent frame.

@jbarry-bdai what would you like to do about this? I'm not super familiar with what the initial issue with the frames was that we had implemented the workaround for but everything appears to work fine besides `link_wr1` being a child of `body`

Screenshot from 2024-03-13 11-39-58

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2684,32 +2684,13 @@ def populate_camera_static_transforms(self, image_data: image_pb2.Image) -> None
...
         for frame_name in image_data.shot.transforms_snapshot.child_to_parent_edge_map:
             if frame_name in excluded_frames:
                 continue
 
             transform = image_data.shot.transforms_snapshot.child_to_parent_edge_map.get(frame_name)
             parent_frame = transform.parent_frame_name
 
-            # special case handling of parent frame to sync with robot state naming, see above
-            if parent_frame == "arm0.link_wr1":
-                parent_frame = "link_wr1"
```
jbarry-bdai, line 2711: Sadly it appears that BD did NOT fix this issue.  This now needs to be "arm0_link_wr1".  Everything else should actually remain.  Sigh.

khughes-bdai:
@jbarry-bdai looking into adding that to this PR as Tiffany is OOO for the rest of this week

jschornak-bdai:
> We currently have THREE arm frames in rviz: arm_link_wr1, arm.link_wr1, and link_wr1. We should figure out where they're all coming from and consolidate down to arm_link_wr1. We will also need to fix at least the hand frame hint in SpotBasic, which uses link_wr1.

I think the intended outcome is for the transform chain for the hand cameras to be independent of the forward kinematics for the arm links. `arm_link_wr1` should be from arm forward kinematics calculated by the RobotStatePublisher (the standalone node in the driver launch file), while `link_wr1` should be from the image publisher node.

One thing to try would be running the Spot state publisher on its own (one way to do that is just by commenting out everything else in the launch file) and see if `arm.link_wr1` is published from there. If so we need to come up with a small fix to filter it out.

> I don't think we should be doing this in Python at all any more? Isn't this in C++ now?

We kept the Python-based image publisher functionality as an optional disabled-by-default feature in the driver node. I'd be in favor of fully deprecating it if we're happy with the C++-based image publisher node.

khughes-bdai:
@jschornak-bdai `arm0.link_wr1` and `link_wr1` are being published in `getTf` from `robot_state.cpp`. I just pushed a change to this PR where the only one of these frames that gets published is `arm_link_wr1`. Should we also leave in `link_wr1`? Pretty sure `arm0.link_wr1` should be taken out as it's not used for anything else

khughes-bdai:
This is the current status of this PR:
* `arm_link_wr1`: stays
* `arm0.link_wr1`: taken out and replaced by `arm_link_wr1`
*  `link_wr1`: taken out and replaced by `arm_link_wr1`

the only frame that gets published anymore is `arm_link_wr1`. The camera sensors that were previously children of `link_wr1` are now children of `arm_link_wr1`. Additionally the `arm_link_fgr` frame stays as a child of `arm_link_wr1`.

Current TF tree:

khughes-bdai (approved):
Tested on robot and the inverse kinematics example still works with the updated frame, additionally the TF tree looks good (see comments)

tcappellari-bdai:
> I should have put this in the issue but here's the specific test to run to make sure that they really did fix this issue:
> 
> 1. Pull up rviz and visualize the point cloud from spot's hand in the vision frame
> 2. Move the hand around and ensure that the point cloud does not move around in the world and that link_wr1 tracks the hand correctly
> 3. Print the list of static transforms this is creating and ensure that there is a transform from link_wr1 to the camera frames and _not_ a link from body to link_wr1.  If there is that link then we need to add link_wr1 to excluded frames.

@jbarry-bdai I think this has been addressed. Can you mark it as done?

jbarry-bdai (approved):
Woohoo!

I wonder why `Kepler/hand` is a child of `Kepler/body` but that's a puzzle for a different time.

PR #298, opened 2024-03-07
Synchronize Spot leasing

mhidalgo-bdai:
## Change Overview

`SpotWrapper` is not thread-safe, and thus reentrant ROS 2 callbacks risk resource contention. This patch synchronizes `SpotWrapper` use in general, and lease management in particular, by making all ROS 2 callbacks mutually exclusive.

It also makes use of https://github.com/bdaiinstitute/spot_wrapper/pull/92 to avoid reaching out to `SpotWrapper` internals.

## Testing Done

In addition to green CI, I manually ran multiple examples on a real Spot (those in this repository and others in downstream repositories as well).

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -39,7 +39,6 @@ from bosdyn.api.spot import robot_command_pb2 as spot_command_pb2
 from bosdyn.api.spot.choreography_sequence_pb2 import Animation, ChoreographySequence
 from bosdyn.client import math_helpers
 from bosdyn.client.exceptions import InternalServerError
-from bosdyn.client.lease import LeaseKeepAlive
 from bosdyn_api_msgs.math_helpers import bosdyn_localization_to_pose_msg
 from bosdyn_msgs.conversions import convert
 from bosdyn_msgs.msg import (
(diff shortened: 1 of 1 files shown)


spot_driver/spot_driver/spot_ros2.py
```diff
@@ -220,7 +218,7 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
         self.callbacks["lease"] = self.lease_callback
         self.callbacks["world_objects"] = self.world_objects_callback
 
-        self.group: CallbackGroup = ReentrantCallbackGroup()
+        self.group: CallbackGroup = MutuallyExclusiveCallbackGroup()
```
mhidalgo-bdai, line 221: I have not thoroughly analyzed the implications of this change yet. If this is too conservative, we may have to synchronize closer to `SpotWrapper` call sites.

mhidalgo-bdai: Circling back to this. I experienced some hangs during testing, which went away after applying https://github.com/bdaiinstitute/spot_ros2/pull/298/commits/7be47914f989c81818d99bd6322d71d733e8d118, which removes callbacks that need not share a callback group from the main, now mutually exclusive, callback group.

mhidalgo-bdai:
@amessing-bdai PTAL

amessing-bdai (approved):

PR #299, opened 2024-03-07
Trajectory batching

gremigi-bdai:
## Change Overview

## Testing Done

### DIFF
--- a/spot_driver/test/pytests/test_tmp.py
+++ b/spot_driver/test/pytests/test_tmp.py
@@ -1,0 +1,350 @@
+"""
+Working with trajectories.
+"""
+
+# pylint: disable=E1101
+
+import logging
+import math
+import time
+from typing import Any, Callable, List, Tuple
+
+from bosdyn.api import (
+    arm_command_pb2,
+    basic_command_pb2,
+    gripper_command_pb2,
+    mobility_command_pb2,
+    robot_command_pb2,
+    synchronized_command_pb2,
+    trajectory_pb2,
+)
+from bosdyn.client.math_helpers import Quat, SE2Pose, SE2Velocity, SE3Pose, SE3Velocity
+from bosdyn.util import duration_to_seconds, seconds_to_duration, seconds_to_timestamp
+
+ContinuousTrajectory1D = Callable[[float], float]
+ContinuousTrajectory2D = Callable[[float], Tuple[SE2Pose, SE2Velocity]]
+ContinuousTrajectory3D = Callable[[float], Tuple[SE3Pose, SE3Velocity]]
+
+logging.basicConfig(level=logging.INFO)
+
+###############################################################################
+# CONTINOUS TRAJECTORIES
+# Here we define some continuous functions to be used to generate
+# sample trajectories.
+
+
+def continuous_trajectory_1d(t: float) -> float:
+    """
+    Given a time t in the trajectory, return a scalar representing a value
+    in the trajectory.
+    """
+    radius = 1.0
+    period = 2.0
+    x = radius * math.cos(2 * math.pi * t / period)
+    return x
+
+
+def continuous_trajectory_2d(t: float) -> Tuple[SE2Pose, SE2Velocity]:
+    """
+    Given a time t in the trajectory, return the SE3Pose and SE3Velocity at
+    this point in the trajectory.
+    """
+
+    # Draw a circle
+    radius = 0.25  # Circle radius in meters
+    period = 2.0  # Time required to go all the way around circle in seconds.
+    x = radius * math.cos(2 * math.pi * t / period)
+    y = radius * math.sin(2 * math.pi * t / period)
+    angle = 0.0
+    vx = -radius * 2 * math.pi / period * math.sin(2 * math.pi * t / period)
+    vy = radius * 2 * math.pi / period * math.cos(2 * math.pi * t / period)
+    va = 0
+    return SE2Pose(x, y, angle), SE2Velocity(vx, vy, va)
+
+
+def continuous_trajectory_3d(t: float) -> Tuple[SE3Pose, SE3Velocity]:
+    """
+    Given a time t in the trajectory, return the SE3Pose and SE3Velocity at
+    this point in the trajectory.
+    """
+
+    # Draw a circle
+    radius = 0.25  # Circle radius in meters
+    period = 2.0  # Time required to go all the way around circle in seconds.
+    x = radius * math.cos(2 * math.pi * t / period)
+    y = radius * math.sin(2 * math.pi * t / period)
+    z = 0.0
+    quat = Quat(1, 0, 0, 0)
+    vx = -radius * 2 * math.pi / period * math.sin(2 * math.pi * t / period)
+    vy = radius * 2 * math.pi / period * math.cos(2 * math.pi * t / period)
+    vz = 0.0
+    return SE3Pose(x, y, z, quat), SE3Velocity(vx, vy, vz, 0, 0, 0)
+
+
+###############################################################################
+# DISCRETE TRAJECTORIES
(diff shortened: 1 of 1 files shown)


spot_driver/spot_driver/spot_ros2.py
```diff
@@ -201,6 +203,21 @@ def set_node_parameter_from_parameter_list(
 class SpotROS(Node):
     """Parent class for using the wrapper.  Defines all callbacks and keeps the wrapper alive"""
 
+    """
```
bhung-bdai, line 206: While I like this method, I wonder if it's a good idea to expose these parameters in the  SpotROS2 constructor

spot_driver/spot_driver/robot_command_util.py
```diff
@@ -0,0 +1,267 @@
...
+        command: The command to inspect.
+
+    Returns:
+        The minimum time_since_reference of all trajectories in the command.
+    """
+    min_time_since_reference = float("inf")
+    if command.HasField("synchronized_command"):
+        if command.synchronized_command.HasField("mobility_command"):
+            mobility_request = command.synchronized_command.mobility_command
+            if trajectory := mobility_request.se2_trajectory_request.trajectory:
```
bhung-bdai, line 193: Are we sure we want to be using this `:=` assignment operator inside the conditional? I thought this would simply always return a `True` because it's just outputting the result of the trajectory assignment, as there's no conditional attached to it.

bhung-bdai (commented):
A few questions but overall looks good. Is there a plan to test the arm impedance control?

spot_driver/spot_driver/robot_command_util.py
```diff
@@ -0,0 +1,267 @@
...
+"""
+Utility class with methods to manipulate robot commands.
+"""
+from typing import Any, List
+
+from bosdyn.api import robot_command_pb2
+from bosdyn.util import duration_to_seconds
+
+
+def is_batch_required(command: robot_command_pb2.RobotCommand, batch_size: int) -> bool:
```
jbarry-bdai, line 12: Nit: This both checks if batching is required and if it's allowed.  Maybe `should_batch`?

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2062,36 +2079,63 @@ def _get_robot_command_feedback(self, goal_id: Optional[str]) -> RobotCommandFee
...
         feedback: Optional[RobotCommandFeedback] = None
         feedback_msg: Optional[RobotCommand.Feedback] = None
+
+        start_time = time.time()
+        time_to_send_command = 0.0
+
+        index = 0
         while (
             rclpy.ok()
+            and goal_handle.is_active
```
jbarry-bdai, line 2116: I'm a little confused how this while loop keeps going until the commands are done sending.  Could you add a comment?

jbarry-bdai (commented):
This is epic.

Would you mind detailing the tests you did?  I suspect they're in the tests and examples file but I'm lazy :-p

spot_driver/spot_driver/spot_ros2.py
gremigi-bdai: It basically keeps going until one of the trajectory command in the batch succeeds i.e. reaches the goal. Because we send the next batch before the previous one is completed, the only batch that succeeds is the last to be sent.
I will add a comment in the code.

spot_driver/spot_driver/robot_command_util.py
gremigi-bdai: In the if condition you highlighted, first the `mobility_request.se2_trajectory_request.trajectory` is evaluated. If it is true (i.e. is not None), its content is assigned to the local variable `trajectory` and the code inside the if condition is executed.
If the condition was always true, even in the scenario where the trajectory is empty, trying to access `points[0]` will throw an exception.

spot_driver/spot_driver/robot_command_util.py
gremigi-bdai: Renamed.

spot_driver/spot_driver/spot_ros2.py
gremigi-bdai: Sorry, I don't understand this comment.

gremigi-bdai: Do you mean these parameters here?
    TRAJECTORY_BATCH_SIZE = 100
    TRAJECTORY_BATCH_OVERLAPPING_POINTS = 20

gremigi-bdai:
> This is epic.
> 
> Would you mind detailing the tests you did? I suspect they're in the tests and examples file but I'm lazy :-p

Yes, there are few unit tests and an example.

In the unit tests I basically verify that all the methods I wrote return the expected values. E.g. the correct number of batches, the correct duration of each batch.

In the example, I send a command with two long trajectories, an arm cartesian trajectory + a body mobility command. The whole command takes 40 seconds to complete and the two trajectories are broken automatically into roughly 10 batches, each batch with a trajectory of 100 points.

When I run the example I observe the robot movements and visually check for jerks, discontinuity or delays during the execution of the two trajectory simultaneously.

Here the output of the example. Originally I moved the robot back and forward, but I was always tripping over the dock, so I decided to rotate the robot body in place instead.

[2.webm](https://github.com/bdaiinstitute/spot_ros2/assets/146826329/4439357f-4cae-40c5-8f99-7ff9efeb9095)

spot_examples/spot_examples/batch_trajectory.py
```diff
@@ -0,0 +1,486 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2023 Boston Dynamics AI Institute LLC. All rights reserved.
+
+"""
+This example illustrates the methodology for dispatching extended-duration
```
khughes-bdai, line 6: Can you add to the README of `spot_examples` to mention that this example has been added?

gremigi-bdai: Added.

khughes-bdai: Thanks! Do we also need to add it to `spot_examples/setup.py`, or is it meant to be run just via python?

spot_driver/spot_driver/spot_ros2.py
gremigi-bdai: Done. I converted the constants into ROS parameters with default values.

spot_examples/spot_examples/batch_trajectory.py
gremigi-bdai: Good catch. Added.

spot_examples/spot_examples/batch_trajectory.py
```diff
@@ -0,0 +1,573 @@
...
+
+        command = self._build_robot_command(
+            arm_cartesian_trajectory=arm_cartesian_trajectory,
+            # arm_joint_trajectory=arm_joint_trajectory,
+            mobility_trajectory=mobility_trajectory,
+            # gripper_trajectory=gripper_trajectory,
+        )
+        action_goal = RobotCommandAction.Goal()
+        convert(command, action_goal.command)
+        self._robot_command_client.send_goal_and_wait("move_arm", goal=action_goal)
```
khughes-bdai, line 534: nit: both the arm and the body are moving in this example, should this be more descriptive than just `move_arm`? Maybe `long_trajectory`?

khughes-bdai (approved):
Just tested this example on robot, everything worked good!

spot_examples/spot_examples/batch_trajectory.py
gremigi-bdai: Done.

tcappellari-bdai (approved):
lgtm

PR #193, opened 2023-11-16
change image requests to greyscale on getting pixel format unsupporte…

skpawar1305:
…d error

handles https://github.com/bdaiinstitute/spot_ros2/issues/189 automatically

PR #303, opened 2024-03-18
spot_image_publisher: Fix pixel format for non RGB fisheye

berndgassmann:
## Change Overview

spot_image_publisher missed to implement the parameter has_rgb_cameras. Now it considers the parameter setting and sets the pixel format to greyscale if has_rgb_cameras=false.

## Testing Done

Now the code runs with our Spot and delivers image data via ROS2. Before the driver complained on invalid pixel format selected, even if has_rgb_cameras was set \c False.


khughes-bdai (approved):
Thank you for your PR! I just tested this on one of our Spots with an RGB camera and the behavior is the same. If this fixes things for non-RGB Spots, it looks good to me.

berndgassmann:
Thanks for fixing the code format. Should have read through your formatting guidelines.

bhung-bdai (approved):

skpawar1305:
this will probably throw error for greyscale with arm as discussed in

https://github.com/bdaiinstitute/spot_ros2/pull/193

Issue #302, opened 2024-03-18
Issue with odom frame becoming child of body in tf


Rodhaaret:
Hey everyone

We just updated the Spot SDK to 4.0.0 and the same with the Spot_ros2 wrapper and are currently trying to get our software up and running correctly again. However we are seeing some issue with the /tf caused by the odometry. 

When looking at the tf tree we see that the odom becomes a child of the body frame, which we assume is the issue, since this was an issue from a previous post: #62 . I assume that the problem occurs when the state_publisher was converted to C++ but not completely sure where to look.

We noticed that the spot_driver gives the following error msg when running the code that previously worked: 
[component_container_mt-7] [ERROR] [1710771142.051852545] [register_node_back]: TF2 exception:
[component_container_mt-7] The tf tree is invalid because it contains a loop.
[component_container_mt-7] Frame base_link exists with parent body.
[component_container_mt-7] Frame body exists with parent odom.
[component_container_mt-7] Frame front_rail exists with parent body.
[component_container_mt-7] Frame rear_rail exists with parent body.
[component_container_mt-7] Frame back exists with parent head.
[component_container_mt-7] Frame head exists with parent body.
[component_container_mt-7] Frame back_fisheye exists with parent back.
[component_container_mt-7] Frame frontleft exists with parent head.
[component_container_mt-7] Frame frontleft_fisheye exists with parent frontleft.
[component_container_mt-7] Frame frontright_fisheye exists with parent frontright.
[component_container_mt-7] Frame frontright exists with parent head.
[component_container_mt-7] Frame left exists with parent head.
[component_container_mt-7] Frame left_fisheye exists with parent left.
[component_container_mt-7] Frame right exists with parent head.
[component_container_mt-7] Frame right_fisheye exists with parent right.
[component_container_mt-7] Frame flat_body exists with parent body.
[component_container_mt-7] Frame odom exists with parent map.
[component_container_mt-7] Frame gpe exists with parent odom.
[component_container_mt-7] Frame vision exists with parent body.
[component_container_mt-7] Frame front_left_hip exists with parent body.
[component_container_mt-7] Frame front_left_upper_leg exists with parent front_left_hip.
[component_container_mt-7] Frame front_left_lower_leg exists with parent front_left_upper_leg.
[component_container_mt-7] Frame front_right_hip exists with parent body.
[component_container_mt-7] Frame front_right_upper_leg exists with parent front_right_hip.
[...]

PR #304, opened 2024-03-18
[SW-859] Resolve TF errors with odom if driver is launched without a namespace

khughes-bdai:
## Change Overview

Currently, if the spot driver is launched without a namespace, the odom frame will incorrectly be a child of the body frame. This is because there is some logic to set the full odom frame ID as `spot_name + "/" + preferred_odom_frame`.  
So with a namespace, this frame is `SpotName/odom`, and without a namespace it becomes `/odom` when it should be `odom` -- leading to some string comparison mismatches in `robot_state.cpp`. This PR makes the TF tree structure the same regardless of if there is a namespace or not. 

Relevant github issue: https://github.com/bdaiinstitute/spot_ros2/issues/302

## Testing Done

- [x] Driver can be launched without a namespace, and the body frame is a child of the odom frame by default. 

- [x] No change to the previous TF structure if the driver is launched with a namespace.

### DIFF
--- a/spot_driver/src/robot_state/state_publisher.cpp
+++ b/spot_driver/src/robot_state/state_publisher.cpp
@@ -35,8 +35,8 @@ StatePublisher::StatePublisher(const std::shared_ptr<StateClientInterface>& stat
 
   const auto preferred_odom_frame = parameter_interface_->getPreferredOdomFrame();
   is_using_vision_ = preferred_odom_frame == "vision";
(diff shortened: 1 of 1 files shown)


khughes-bdai:
* **#304** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/304?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

bhung-bdai (approved):

tcappellari-bdai (approved):
nice catch

PR #313, opened 2024-03-21
Switch ROS2 QoS to most compatible values for publishers

berndgassmann:
The QoS settings of the publisher are set to the most compatible values, so that theoretically any client configuration should work. Like this, the clients decide on actually selected mode.

On the clients (rviz, battery-monitor) on the other hand it is switched to SensorData defaults since it is not relevant to receive each and every package on the status data. If something is lost, it's rather better to receive the latest greatest one instead of enforcing resending of outdated data.


jschornak-bdai (commented):
Could you disentangle the QoS changes for the compressed image publisher changes that are now in #316?

@berndgassmann which DDS middleware are you using? I did originally try setting the image publisher QoS to BestEffort, since I agree that this makes more sense for a continuously-publishing sensor topic, but counterintuitively that made my framerate _much worse_ instead of better like I would expect. This was using FastDDS. I haven't deeply explored why specifically that could be happening, though, and it could very well be an issue specific to my local environment. Certainly worth investigating this, though.

---

> the people who have set their receivers to Reliable have to change their receiver reliability: at least with the newer ROS2 environments you immediately get warnings from the middleware on that conflict.

I wish subscribers were implemented to just pick the correct reliability QoS setting based on what the publisher is using. IMO it doesn't make sense to have this configurable on both the publisher and subscriber side of things, since ultimately it's the subscriber that's determining the data that gets transmitted on the topic. Nothing to do about that here of course :(

berndgassmann:
Hmm, yes you are right in all the points.
My colleague just observed the same issue with the frame rate and QoS settings on the realsense cameras we are heaviliy using. I am using ros2-humble with FastDDS currently on Ubuntu 22.04. But testing out CycloneDDS made things even worse. I am also a bit confused on the extremely bad behaviour of both the two DDS implementatons. My local machine is doing nearly nothing i9-13900K. And the middleware is not able to transport 5 image streams in real-time from spot-IP interface to ros2-middleware. So it seems that at least all users of ROS2 middleware have similar issue. Maybe there are some hidden options that remove some sleeps in the ROS2/DDS stack (just kidding).
I am sooo close to throw away both Midelewares and exchange them by our own (10 years old) shared memory middleware we applied in the automotive space using the CommonAPI that time. Then the speed would be real-time, the CPU would have to do nothing and at least if the network driver is enabled to put it's data via DMA directly into the shared memory blob that is sent out -- even no memory copies of the mass-data are required at all. I know that such is possible as we did that for Audio Video Bridging in Time Sensitive Networks.
Coming back from the digression: Maybe the issues we all observe have nothing to do with DDS inherent capabilities, but rather the rings of onion around the core thing we want to achieve. We have now:
- Network reception (dma driver buffer) -> SPOT API (data copy? hopefully not) -> spot_ros2 (data copy into sensor_msgs::msg::Image) -> DDS (additional data copy when sending out ?)
I'd guess most of the time is lost because of data copies, but I didn't look into that level of detail yet. At least, in an ideal case one would:
- first allocate a data buffer at DDS side which is used for sending out the data via ROS (without copy); in a local setup (ROS2 sender/receiver are on the same host machine) this should directly return a shared memory buffer pointer; If looking into ROS2 FastDDS a bit more, this should be possible (https://github.com/eProsima/Fast-DDS/discussions/2550) 
- then pass this data pointer to the bodsdyn::api::whatever this uses to capture the incoming data [that might be required as additional interface at the driver side]
I am pretty sure that especially for a robot like SPOT having many video streams, and depth sensor stream in addition this would make totally sense. I know that our code generator in the old automotive CommonAPI times did that under the hood automatically for the user, because we also defined the processing interfaces more strict (making it easier for the user, but for sure less flexible).

I'll remove the QoS changes as you suggest. When you like we can have a detailed chat on the above ideas ... maybe in a different form than PR comments ;-). I guess a lot of people would be glad of having a lean interface especially when working with embedded low-power compute as it's required when using autonomous robots with legs...

berndgassmann:
first step: removed the merges of the other branch to keep both PRs more independent (hope this works).

berndgassmann:
second step done: merge latest main in here
third step to come...

berndgassmann:
looks as if the humble container build misses some access rights...

berndgassmann:
I updated the PR title and the PR description to what remained now in the PR.

spot_driver/src/robot_state/state_middleware_handle.cpp
```diff
@@ -35,8 +35,13 @@ constexpr auto kEndEffectorForceTopic{"status/end_effector_force"};
 constexpr auto kManipulatorTopic{"manipulation_state"};
 
 rclcpp::QoS makeQoS() {
```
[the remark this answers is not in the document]
berndgassmann: Thanks, you are so right. See the commit solving this (and fixing the image publisher missed on merging the changes from main)

bhung-bdai (approved):
Tested on hardware with Rviz. It seems to accurately represent the different (Rviz uses Reliable, but the driver uses best effort)

khughes-bdai:
This also looks good to me, but want to check in with @tcappellari-bdai as I thought she found some sort of mismatch with `depth_image_proc` with these updated QoS's?

tcappellari-bdai:
> Tested on hardware with Rviz. It seems to accurately represent the different (Rviz uses Reliable, but the driver uses best effort)

@bhung-bdai were you getting QoS mismatch errors from all the `camera/images` and `camera/camera_info` topics?

bhung-bdai:
I hadn't noticed it @tcappellari-bdai, but I wasn't specifically paying attention to know what to look for.

spot_driver/include/spot_driver/api/middleware_handle_base.hpp
```diff
@@ -0,0 +1,22 @@
...
+ * @brief Base class for common ROS2 publisher definitions
+ */
+class MiddlewareHandleBase {
+ public:
+  rclcpp::QoS makePublisherQoS(size_t const publisherHistoryDepth) const {
+    // most compatible publisher durability: transient local
+    // most compatible publisher reliabilty: reliable
+    // see also
+    // https://docs.ros.org/en/iron/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-compatibilities
+    return rclcpp::QoS(publisherHistoryDepth).transient_local().reliable();
```
tcappellari-bdai, line 19: @bhung-bdai @khughes-bdai did you guys change the QoS to best effort? Just realized I forgot to do that before testing again lol and the errors returned

tcappellari-bdai: I get a bunch of mismatched QoS from the nodes we generate withe `image_pipeline`'s `depth_proc_node`s. I had looked into their source code with Mich last time and saw they set their QoS's all to ROS default (reliable)

```
[component_container_mt-8] [WARN] [1715004057.643811002] [Opal.register_node_right]: New publisher discovered on topic '/Opal/depth/right/image', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[component_container_mt-8] [WARN] [1715004057.644263948] [Opal.register_node_right]: New publisher discovered on topic '/Opal/depth/right/camera_info', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[component_container_mt-8] [WARN] [1715004057.644683021] [Opal.point_cloud_xyzrgb_node_hand]: New publisher discovered on topic '/Opal/camera/hand/image', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[component_container_mt-8] [WARN] [1715004057.644960723] [Opal.point_cloud_xyzrgb_node_hand]: New publisher discovered on topic '/Opal/camera/hand/camera_info', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[component_container_mt-8] [WARN] [1715004057.644978954] [Opal.register_node_hand]: New publisher discovered on topic '/Opal/camera/hand/camera_info', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[component_container_mt-8] [WARN] [1715004057.645372683] [Opal.register_node_hand]: New publisher discovered on topic '/Opal/depth/hand/image', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[component_container_mt-8] [WARN] [1715004057.645755978] [Opal.register_node_hand]: New publisher discovered on topic '/Opal/depth/hand/camera_info', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[spot_image_publisher_node-2] [WARN] [1715004057.646131217] [Opal.image_publisher]: New subscription discovered on topic '/Opal/camera/back/image', requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[spot_image_publisher_node-2] [WARN] [1715004057.646303530] [Opal.image_publisher]: New subscription discovered on topic '/Opal/camera/back/camera_info', requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[spot_image_publisher_node-2] [WARN] [1715004057.646323736] [Opal.image_publisher]: New subscription discovered on topic '/Opal/depth/back/image', requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[spot_image_publisher_node-2] [WARN] [1715004057.646337799] [Opal.image_publisher]: New subscription discovered on topic '/Opal/depth/back/camera_info', requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[spot_image_publisher_node-2] [WARN] [1715004057.646349650] [Opal.image_publisher]: New subscription discovered on topic '/Opal/camera/frontleft/image', requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[spot_image_publisher_node-2] [WARN] [1715004057.646361215] [Opal.image_publisher]: New subscription discovered on topic '/Opal/camera/frontleft/camera_info', requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
```

tcappellari-bdai (approved):
I'm going to approve this and merge it in but leave the QoS as reliable for now while we debug the other issues. Thanks @berndgassmann for your contribution!

PR #327, opened 2024-04-01
Explicitly define interface class copy and move constructors and mark functions as nodiscard

jschornak-bdai:
## Change Overview

Fixes some things the clangd plugin in my IDE was complaining about:
- Mark member functions in derived classes which return values as being `[[nodiscard]]`. This helps enforce correct usage.
  - Related: fix one place where a return value was being ignored.
- For the base interface classes, delete the copy constructor and explicitly define the default move constructor to indicate that these classes are move-only. This is primarily to enforce a particular pattern of usage, since it matches the way the implementations of these classes are used in the nodes.
- Fix minor doc typos and unused includes.

## Testing Done

If the PR branch compiles and tests pass, then this is all set.


amessing-bdai (approved):

khughes-bdai (approved):
LGTM

PR #335, opened 2024-04-04
Impedance example for trajectory batching.

gremigi-bdai:
## This is an example to batch trajectory in an impedance command.

To execute an impedance trajectory you must comment out some lines 601 - 607. I added comments in the code to explain what is working, what is not working and how to run an impedance trajectory instead of a cartesian trajectory.


khughes-bdai (approved):
LGTM, default state (just arm_cartesian_trajectory) worked on robot, and uncommenting the lines to set the arm_impedance_trajectory also worked smoothly

tcappellari-bdai (approved):

PR #312, opened 2024-03-21
[PN-1] Add ObjectSynchronizer to add frames broadcast to TF as WorldObjects in Spot's internal world model

jschornak-bdai:
## Change Overview

- Adds a new ObjectSynchronizer node that monitors ROS's TF topics and the Spot's internal world representation and creates new Spot WorldObjects corresponding to TF frames that are published by sources outside the Spot driver.
  - This node assumes responsibility for publishing TF frames for other Spot WorldObjects to ROS's TF topics, so the equivalent functionality in the Python node has been removed.

## Testing Done

- [ ] Use `ros2 run tf2_ros static_transform_publisher` to publish a new transform to TF. This will result in a corresponding WorldObject being added to Spot's internal world representation. Other Spot API tools that can plan relative to WorldObjects will be able to use this new object as a goal.
- [ ] Ensure that the TF frames for docks, AR tags, and other WorldObjects which are created automatically by Spot are still published.

PR #337, opened 2024-04-04
Set that RclcppTfListenerInterface's TF buffer is using a separate thread to monitor TF data

jschornak-bdai:
## Change Overview

There is a noisy warning message from the TF buffer cautioning that `lookupTransform` should not be called with a timeout unless a separate thread is used to listen to TF data. We are indeed using a separate thread (since we require that RclcppTfListenerInterface be spun by a multithreaded executor) and we're not using the timeout duration when looking up transforms, so we can set that the buffer is using a separate thread to quiet this warning.

This also fixes an incorrect copyright year for a file added in #312.

## Testing Done

- [x] Run driver nodes, confirm absence of warning message.

### DIFF
--- a/spot_driver/include/spot_driver/interfaces/rclcpp_tf_listener_interface.hpp
+++ b/spot_driver/include/spot_driver/interfaces/rclcpp_tf_listener_interface.hpp
@@ -16,6 +16,7 @@
 namespace spot_ros2 {
 /**
  * @brief Implements TfListenerInterfaceBase to use the rclcpp TF system.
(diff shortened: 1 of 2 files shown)


tcappellari-bdai (approved):
lgtm

khughes-bdai (approved):
LGTM, I can also test this if you haven't yet/aren't in office -- let me know!

khughes-bdai:
Going to test this on-robot right now to make sure that everything still works

PR #342, opened 2024-04-11
Special handling of the hand camera format such that it always has RGB format

philipsgithub:
We assume the hand camera to always have rgb format, independent of the parameter "has_rgb_cameras"

We tested it on our spot, which has grayscale body cameras, tests on an rgb-spot would be required.

This PR is related to https://github.com/bdaiinstitute/spot_ros2/pull/193 and slightly adapts the simplified approach mentioned there.


philipsgithub:
> Just tested on an RGB Spot, and it looks good! Would you mind adding a quick comment just above this line that mentions that the hand camera is always RGB?

good to hear. i hope the comment is expressive enough

khughes-bdai (approved):

PR #347, opened 2024-04-12
[SW-947] Add support for custom port and certificate

mhidalgo-bdai:
## Change Overview

Precisely what the title says. This patch restores tests after https://github.com/bdaiinstitute/spot_wrapper/pull/103 and opens the door to (`launch_`)`pytest` driven C++ integration tests, as both PRs combined effectively enable the Spot ROS 2 driver to run against a mock Spot over secure gRPC channels.

## Testing Done

- [x] Integrations tests passed
- [x] Examples ran with no issues

PR #348, opened 2024-04-12
[SW-958] hotfix for TF spew from driver

khughes-bdai:
## Change Overview

The frame snapshot obtained when getting apriltag world objects from spot contains some static frames like the camera frames. This results in extreme spew from the driver about TF extrapolation to the future whenever an apriltag is in view which affects camera and point cloud publishing. (introduced in https://github.com/bdaiinstitute/spot_ros2/pull/312)

This solution filters out the frames in the world object TF broadcaster so that it only broadcasts the world object frames and doesn't rebroadcast the static frames. 

In working on this, I uncovered another issue with getting the list of current frames within the SyncWorldObjects function. The problem is the list of TF frames includes old TF frames that correspond to tags the robot can't see anymore (see: https://github.com/bdaiinstitute/spot_ros2/blob/d48450ec6fc670d88f92d6a8a88b65b296271b06/spot_driver/src/interfaces/rclcpp_tf_listener_interface.cpp#L18), leading to a bunch of extrapolation to the past errors if the robot initially could see a tag but now cannot. As this is a new feature, I think this can be temporarily disabled to get `main` functional again. 

## Testing Done

- [x] launched driver robot in view of an apriltag, spew is gone. Can move the robot out of view, then back in view of the apriltags/docks, still no spew. 
- [x] all `spot_examples` work with no spew

### DIFF
--- a/spot_driver/include/spot_driver/conversions/robot_state.hpp
+++ b/spot_driver/include/spot_driver/conversions/robot_state.hpp
@@ -116,19 +116,15 @@ std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::RobotState& r
  * @param prefix  The prefix to apply to all robot frame IDs. This corresponds to the name of the robot. It is expected
  * to terminate with `/`.
  * @param preferred_base_frame_id Frame ID to use as the base frame of the TF tree. Must be either "odom" or "vision".
+ * @param frames_to_ignore Set of frames to not include in the TF tree.
  * @return If the input frame tree snapshot contains a non-zero number of entries, return a TFMessage containing
  * this data. Otherwise, return nullopt.
  */
 std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::FrameTreeSnapshot& frame_tree_snapshot,
                                               const google::protobuf::Timestamp& timestamp_robot,
                                               const google::protobuf::Duration& clock_skew, const std::string& prefix,
-                                              const std::string& preferred_base_frame_id);
-
-std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::FrameTreeSnapshot& frame_tree_snapshot,
-                                              const google::protobuf::Timestamp& timestamp_robot,
-                                              const google::protobuf::Duration& clock_skew, const std::string& prefix,
                                               const std::string& preferred_base_frame_id,
-                                              std::set<std::string, std::less<>> frames_to_ignore);
+                                              std::set<std::string, std::less<>> frames_to_ignore = {});
 /**
  * @brief Create an TwistWithCovarianceStamped ROS message representing Spot's body velocity by parsing a RobotState
  * message.
--- a/spot_driver/include/spot_driver/conversions/robot_state.hpp
+++ b/spot_driver/include/spot_driver/conversions/robot_state.hpp
@@ -124,7 +124,7 @@ std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::FrameTreeSnap
                                               const google::protobuf::Timestamp& timestamp_robot,
                                               const google::protobuf::Duration& clock_skew, const std::string& prefix,
                                               const std::string& preferred_base_frame_id,
-                                              std::set<std::string, std::less<>> frames_to_ignore = {});
+                                              const std::set<std::string, std::less<>> frames_to_ignore = {});
 /**
  * @brief Create an TwistWithCovarianceStamped ROS message representing Spot's body velocity by parsing a RobotState
  * message.
--- a/spot_driver/include/spot_driver/conversions/robot_state.hpp
+++ b/spot_driver/include/spot_driver/conversions/robot_state.hpp
@@ -4,6 +4,7 @@
 
 #include <bosdyn/api/robot_state.pb.h>
 #include <bosdyn_api_msgs/msg/manipulator_state.hpp>
+#include <functional>
 #include <geometry_msgs/msg/twist_with_covariance_stamped.hpp>
 #include <geometry_msgs/msg/vector3_stamped.hpp>
 #include <map>
@@ -12,6 +13,7 @@
 #include <sensor_msgs/msg/camera_info.hpp>
 #include <sensor_msgs/msg/image.hpp>
 #include <sensor_msgs/msg/joint_state.hpp>
+#include <set>
 #include <spot_msgs/msg/battery_state_array.hpp>
 #include <spot_msgs/msg/behavior_fault_state.hpp>
 #include <spot_msgs/msg/e_stop_state_array.hpp>
@@ -122,6 +124,11 @@ std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::FrameTreeSnap
                                               const google::protobuf::Duration& clock_skew, const std::string& prefix,
                                               const std::string& preferred_base_frame_id);
 
+std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::FrameTreeSnapshot& frame_tree_snapshot,
+                                              const google::protobuf::Timestamp& timestamp_robot,
+                                              const google::protobuf::Duration& clock_skew, const std::string& prefix,
+                                              const std::string& preferred_base_frame_id,
+                                              std::set<std::string, std::less<>> frames_to_ignore);
 /**
  * @brief Create an TwistWithCovarianceStamped ROS message representing Spot's body velocity by parsing a RobotState
  * message.
--- a/spot_driver/include/spot_driver/conversions/robot_state.hpp
+++ b/spot_driver/include/spot_driver/conversions/robot_state.hpp
@@ -116,7 +116,7 @@ std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::RobotState& r
  * @param prefix  The prefix to apply to all robot frame IDs. This corresponds to the name of the robot. It is expected
  * to terminate with `/`.
  * @param preferred_base_frame_id Frame ID to use as the base frame of the TF tree. Must be either "odom" or "vision".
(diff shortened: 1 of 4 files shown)


bhung-bdai (commented):
This seems like a weird way to republish these things.can we just avoid adding them in the first place?

khughes-bdai:
adding do not merge since we really should be doing it through here, but the list of static transforms is never updated.  
https://github.com/bdaiinstitute/spot_ros2/blob/main/spot_driver/src/interfaces/rclcpp_tf_broadcaster_interface.cpp

khughes-bdai:
This is the issue with the `SyncWorldObjects` function, which I've narrowed down to this call to `getAllFrameNames`
https://github.com/bdaiinstitute/spot_ros2/blob/d48450ec6fc670d88f92d6a8a88b65b296271b06/spot_driver/src/object_sync/object_synchronizer.cpp#L400

<div class='graphite__hidden'>
          <div>🎥 Video uploaded on Graphite:</div>
            <a href="https://app.graphite.dev/media/video/Xynj6CBpA3NqqBFfE8Q9/7e4e5b29-a973-4008-a69e-2edc08e3e80e.webm">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/Xynj6CBpA3NqqBFfE8Q9/7e4e5b29-a973-4008-a69e-2edc08e3e80e.webm">fiducials_go_out_of_range.webm</video>


Since there's a test for the `SyncWorldObjects` code function call I commented out, this PR currently won't pass CI. Planning to remove this test for the time being as it's still in the git history once we fix the issue, but open to suggestions.

spot_driver/include/spot_driver/conversions/robot_state.hpp
```diff
@@ -114,14 +116,15 @@ std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::RobotState& r
...
  * @return If the input frame tree snapshot contains a non-zero number of entries, return a TFMessage containing
  * this data. Otherwise, return nullopt.
  */
 std::optional<tf2_msgs::msg::TFMessage> getTf(const ::bosdyn::api::FrameTreeSnapshot& frame_tree_snapshot,
                                               const google::protobuf::Timestamp& timestamp_robot,
                                               const google::protobuf::Duration& clock_skew, const std::string& prefix,
-                                              const std::string& preferred_base_frame_id);
-
+                                              const std::string& preferred_base_frame_id,
+                                              const std::set<std::string, std::less<>> frames_to_ignore = {});
```
amessing-bdai, line 127: Do you need the `std::less<>`?

khughes-bdai: I couldn't get the code to build without it unfortunately 
```
/home/khughes/spot_ros2_ws/src/spot_ros2/spot_driver/src/object_sync/object_synchronizer.cpp:509:34: error: no matching function for call to ‘getTf(const bosdyn::api::FrameTreeSnapshot&, const google::protobuf::Timestamp&, const google::protobuf::Duration&, std::string&, std::string&, const std::set<std::__cxx11::basic_string<char>, std::less<void> >&)’
...
/home/khughes/spot_ros2_ws/src/spot_ros2/spot_driver/include/spot_driver/conversions/robot_state.hpp:127:75: note:   no known conversion for argument 6 from ‘set<[...],less<void>>’ to ‘set<[...],less<std::__cxx11::basic_string<char>>>’
  127 |                                               const std::set<std::string> frames_to_ignore = {});

```

amessing-bdai: Also, should it be a reference?

tcappellari-bdai:
Based on Katie tracing the problem back to the `getAllFrameNames` call returning past frames that no longer exist along with the current ones and some digging around online, I think that might actually just be a feature of `tf2` since you can look up transforms from specific timestamps in the past so I guess that stuff is still stored? Or the very least their built in get all frame names function has a bug that keeps all frame names even if the actual listener and buffer no longer have those frames transforms stored. 

Or maybe we just need to call it more than once to get an updated list?

@khughes-bdai does everything seem to work ok besides the endless tf extrapolation error spew?
If the problem is mainly just reduce spew then maybe we update what this does in the catches
https://github.com/bdaiinstitute/spot_ros2/blob/d48450ec6fc670d88f92d6a8a88b65b296271b06/spot_driver/src/interfaces/rclcpp_tf_listener_interface.cpp#L22-L36

Or potentially this should be in a try catch?
https://github.com/bdaiinstitute/spot_ros2/blob/d48450ec6fc670d88f92d6a8a88b65b296271b06/spot_driver/src/object_sync/object_synchronizer.cpp#L412-L413

I'm brainstorming but these are some ideas ive had

khughes-bdai:
@tcappellari-bdai the problem now is solely the call to `getAllFrameNames`. I'm not actually trying to look up this transform since the loop in `syncWorldObjects` will always break before this call if you aren't adding custom tf frames (ie they are always immutable or internal frames https://github.com/bdaiinstitute/spot_ros2/blob/d48450ec6fc670d88f92d6a8a88b65b296271b06/spot_driver/src/object_sync/object_synchronizer.cpp#L404). It seems to be some bug with the `buffer_` member in `RclcppTfListenerInterface`(https://github.com/bdaiinstitute/spot_ros2/blob/d48450ec6fc670d88f92d6a8a88b65b296271b06/spot_driver/src/interfaces/rclcpp_tf_listener_interface.cpp#L19)

khughes-bdai:
maybe it is possible to re-initialize the buffer every time this function is called. I can look into that and see if it resolves the issue

tcappellari-bdai:
> @tcappellari-bdai the problem now is solely the call to `getAllFrameNames`. I'm not actually trying to look up this transform since the loop in `syncWorldObjects` will always break before this call if you aren't adding custom tf frames (ie they are always immutable or internal frames
> 
> https://github.com/bdaiinstitute/spot_ros2/blob/d48450ec6fc670d88f92d6a8a88b65b296271b06/spot_driver/src/object_sync/object_synchronizer.cpp#L404
> 
> ). It seems to be some bug with the `buffer_` member in `RclcppTfListenerInterface`(
> https://github.com/bdaiinstitute/spot_ros2/blob/d48450ec6fc670d88f92d6a8a88b65b296271b06/spot_driver/src/interfaces/rclcpp_tf_listener_interface.cpp#L19
> 
> ) which i am worried is a TF internal issue

Ya what I'm saying is TF is pretty much built to fail often which is why basically all tutorials have the get transform call inside of some try catch. I think the expectation of using that built in get all frame names function is to use them to call the lookup transform func but expecting that some will fail. Like I think this might just be how TF works (whether we agree with/like it or not).

Otherwise, I dont think the buffer itself needs to be reinitialized. I was asking if we ever call `getAllFrameNames` again after the first time or if thats just The List™ of frames now.

khughes-bdai:
`getAllFrames` gets continuously called in the `syncWorldObjects` timer. I'm wondering if we try to try/catch it if it won't return any frames.

khughes-bdai:
`getAllFrames` gets continuously called in the `syncWorldObjects` timer. I'm wondering if we try to try/catch it if it won't return any frames.

tcappellari-bdai:
> `getAllFrames` gets continuously called in the `syncWorldObjects` timer. I'm wondering if we try to try/catch it if it won't return any frames.

The try catch would just be for the actual lookup so once the lookup function returns the extrapolation into the past error, it should catch it and just continue instead of spewing them out endlessly

khughes-bdai:
discovered another serious instance of spew from the `syncWorldObjects` function which further reinforces my belief that this part of the node should be disabled for now. This was when running the inverse kinematics example on main -- the node is constantly adding new objects that shouldn't be there. 

<div class='graphite__hidden'>
          <div>🎥 Video uploaded on Graphite:</div>
            <a href="https://app.graphite.dev/media/video/Xynj6CBpA3NqqBFfE8Q9/f6bedb8b-c762-4cb5-8462-2f4583682659.webm">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/Xynj6CBpA3NqqBFfE8Q9/f6bedb8b-c762-4cb5-8462-2f4583682659.webm">Screencast from 04-17-2024 04:30:21 PM.webm</video>

khughes-bdai:
discovered another serious instance of spew from the `syncWorldObjects` function which further reinforces my belief that this part of the node should be disabled for now. This was when running the inverse kinematics example on main -- the node is constantly adding new objects that shouldn't be there. 

<div class='graphite__hidden'>
          <div>🎥 Video uploaded on Graphite:</div>
            <a href="https://app.graphite.dev/media/video/Xynj6CBpA3NqqBFfE8Q9/f6bedb8b-c762-4cb5-8462-2f4583682659.webm">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/Xynj6CBpA3NqqBFfE8Q9/f6bedb8b-c762-4cb5-8462-2f4583682659.webm">Screencast from 04-17-2024 04:30:21 PM.webm</video>

mhidalgo-bdai:
I got here from https://github.com/bdaiinstitute/spot_ros2/pull/347. The way I see it, the implementation is problematic because (a) which Spot frames are assumed static and which are not isn't common knowledge across the codebase, and (b) there is no way to clearly identify object frames that do not originate in Spot.

If we define (or compute) the set of Spot frames that are static in advance, there is no need to go about hard-coding all internal frames. There's nothing wrong with re-publishing the odom to body at different times from different nodes. In fact, you get perfect synchronization that way instead of relying on `tf2` interpolation. Furthermore, if we make object registration from ROS 2 explicit, using service calls for example, there is no need to go over all known `tf` frames (some of which may be very old).

tcappellari-bdai (approved):
We will merge this in as a temporary patch for reducing the TF spew and will make a separate ticket for changing the implementation of object sync to avoid this issue entirely.

PR #358, opened 2024-04-23
Handle Greyscale images correctly

berndgassmann:
- Fix: do not decode greyscale camera images as color image
- Pass the (already implemented) option to NOT decompress the images to the user parameter interface as "publish_raw_rgb_cameras" to let the user decide if the decompression should be performed or not; parameter was hard coded as false previously
- increase the publisher's history depth to support slower receivers

## Change Overview

This is the part on fixing the greyscale processing from #316

## Testing Done

The unit test for the parameter input was added in the commit and unit tests run successfully.
No CI-test extensions planned in addition. To actually test the greyscale changes one would require hardware in the loop tests...


khughes-bdai (approved):
LGTM after a test on an RGB Spot :)

tcappellari-bdai (approved):
lgtm

PR #359, opened 2024-04-23
Fix duplicated ros_node creation

berndgassmann:
## Change Overview

If the ImagesMiddlewareHandle is constructed using the node_options, it's creating it's own ros-node.
But when initialized from within SpotImagePublisherNode, the already existing ros-node must be used. Otherwhise it results in two ros-nodes with identical names and corresponding ROS2 warnings.

## Testing Done
Compiles and runs now with only one /image_publisher node ;-)


khughes-bdai (approved):
Thank you for the PR! Just tested and everything looks good on our side.

PR #366, opened 2024-04-26
Fb set arm velocity

MGBla:
## Change Overview

Corresponding Driver calls to the PR on the wrapper Repo [104](https://github.com/bdaiinstitute/spot_wrapper/pull/104) to control the robot using a VelocityCommand interface. I decided to use a topic interface to control the arm since i want to have direct control over it instead of a longer running service call.
The angular and Cartesian velocity were thoroughly tested during operation on our spot. For the cylindrical velocity I performed some basic tests but didn't evaluate over hours as the other two input methods since I don't really use them in this reference frame.
## Testing Done
    - [x] tested that these new services work on robot

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -27,6 +27,7 @@ from bdai_ros2_wrappers.single_goal_multiple_action_servers import (
     SingleGoalMultipleActionServers,
 )
 from bosdyn.api import (
+    arm_command_pb2,
     geometry_pb2,
     gripper_camera_param_pb2,
     image_pb2,
@@ -35,6 +36,7 @@ from bosdyn.api import (
     trajectory_pb2,
     world_object_pb2,
 )
+
 from bosdyn.api.geometry_pb2 import Quaternion, SE2VelocityLimit
 from bosdyn.api.spot import robot_command_pb2 as spot_command_pb2
 from bosdyn.api.spot.choreography_sequence_pb2 import Animation, ChoreographySequence, ChoreographyStatusResponse
@@ -44,6 +46,7 @@ from bosdyn_api_msgs.math_helpers import bosdyn_localization_to_pose_msg
 from bosdyn_msgs.conversions import convert
 from bosdyn_msgs.msg import (
     ArmCommandFeedback,
+    ArmVelocityCommandRequest,
     Camera,
     FullBodyCommandFeedback,
     GripperCommandFeedback,
@@ -495,6 +498,9 @@ class SpotROS(Node):
 
         self.create_subscription(Twist, "cmd_vel", self.cmd_velocity_callback, 1, callback_group=self.group)
         self.create_subscription(Pose, "body_pose", self.body_pose_callback, 1, callback_group=self.group)
+        if has_arm:
+            self.create_subscription(ArmVelocityCommandRequest, "arm_velocity",
+                                     self.set_arm_velocity_callback, 1, callback_group=self.group)
         self.create_service(
             Trigger,
             "claim",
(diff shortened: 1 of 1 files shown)


spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2919,6 +2925,35 @@ def handle_set_gripper_camera_parameters(
 
         return response
 
+    def set_arm_velocity_callback(self, data: ArmVelocityCommandRequest) -> None:
```
tcappellari-bdai, line 2928:
```suggestion
    def handle_arm_velocity_command(self, arm_velocity_command: ArmVelocityCommandRequest) :
```

tcappellari-bdai (commented):
This would be better as a service call instead of a subscriber callback.
There are also some conversion methods we have already provided to convert between protobufs and ROS messages so you don't have to convert them manually.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -495,6 +497,10 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
 
         self.create_subscription(Twist, "cmd_vel", self.cmd_velocity_callback, 1, callback_group=self.group)
         self.create_subscription(Pose, "body_pose", self.body_pose_callback, 1, callback_group=self.group)
+        if has_arm:
+            self.create_subscription(
+                ArmVelocityCommandRequest, "arm_velocity", self.set_arm_velocity_callback, 1, callback_group=self.group
+            )
```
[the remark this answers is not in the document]
khughes-bdai: +1, to keep with the style of the rest of the repo I think it would be best to handle this as either a service or an action. We have methods for sending standard robot commands as either an action (https://github.com/bdaiinstitute/spot_ros2/blob/492587a8962c5796994b633bb451c5e1a86c8906/spot_driver/spot_driver/spot_ros2.py#L2231) or a service (https://github.com/bdaiinstitute/spot_ros2/blob/492587a8962c5796994b633bb451c5e1a86c8906/spot_driver/spot_driver/spot_ros2.py#L2216) -- we use the action implementation for longer commands and the service implementation for faster/"streaming" commands. If you want more direct control you should be able to achieve the same behavior you've written here with a service call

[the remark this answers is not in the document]
MGBla: Hey normally I would agree to keep the whole repo ins the same style but in this case I had specific reasons to disregard that and implement it another way.
The ArmVelocity movement is more related to basic movemnet of the robot or in this case the cmd_vel interface.
Using a service at this point would have too much time overhead which would result in way less smooth movement ( because, yes i first tried it as a service and it was horribly unresponsive). Even though services are kind of fast they still require all the internal handshakes and are not suitable for this kind of control.

spot_driver/spot_driver/spot_ros2.py
MGBla: Yeah sure, why not, however the creation of the subscription needs also to be changed. will fix it

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2919,6 +2929,43 @@ def handle_set_gripper_camera_parameters(
...
+        cartesian_velocity.velocity_in_frame_name.z = (
+            arm_velocity_command.command.cartesian_velocity.velocity_in_frame_name.z
+        )
+        cartesian_velocity.frame_name = arm_velocity_command.command.cartesian_velocity.frame_name
+
+        proto_command = arm_command_pb2.ArmVelocityCommand.Request(
+            cylindrical_velocity=cylindrical_velocity,
+            angular_velocity_of_hand_rt_odom_in_hand=angular_velocity,
+            cartesian_velocity=cartesian_velocity,
+        )
```
khughes-bdai, line 2965: You should be able to replace this with the following, using some built in proto to ros conversion:
```
proto_command = arm_command_pb2.ArmVelocityCommand.Request()
convert(arm_velocity_command, proto_request)
```
Can you verify that this still works with your example?

MGBla: Will test it later this afternoon on the robot :)

khughes-bdai: Hello, just wondering if you had a chance to test this out?

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2919,6 +2929,43 @@ def handle_set_gripper_camera_parameters(
...
+
+        return self.spot_wrapper.spot_arm.handle_arm_velocity(proto_command)
```
khughes-bdai, line 2967:
```suggestion
        return self.spot_wrapper.spot_arm.handle_arm_velocity(proto_command, self.cmd_duration)
```
If we want this to mimic the `cmd_vel` topic, we can use the same duration field as it.

spot_driver/spot_driver/spot_ros2.py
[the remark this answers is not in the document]
khughes-bdai: We discussed this some more and it sounds good to keep it as a subscription to be analogous to `cmd_vel` but for the arm

MGBla:
@corot Sorry have slept a little on this PR. I can test the Version of @kabirkedia next week when I have access to our spot again to see if the reactive teleoperation works this way as well.

PR #371, opened 2024-04-29
Add option to turn off compressed images

tcappellari-bdai:
## Change Overview

Added the flag `publish_compressed_images` to spot driver to turn publishing compressed images on and off.

Also changed to default to publishing only uncompressed images instead of both compressed and uncompressed (for now) 

## Testing Done

Ran driver on RGB spot and checked topics and messages were behaving as expected

### DIFF
--- a/spot_driver/include/spot_driver/interfaces/image_client_interface.hpp
+++ b/spot_driver/include/spot_driver/interfaces/image_client_interface.hpp
@@ -33,6 +33,7 @@ class ImageClientInterface {
   ImageClientInterface& operator=(const ImageClientInterface&) = delete;
   virtual ~ImageClientInterface() = default;
   virtual tl::expected<GetImagesResult, std::string> getImages(::bosdyn::api::GetImageRequest request,
-                                                               bool uncompress_images, bool publish_compressed_images) = 0;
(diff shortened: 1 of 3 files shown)


amessing-bdai:
@tcappellari-bdai This parameter seems misleading. We aren't compressing the image. We receive it from spot as a compressed image and we are choosing whether to publish it or not. 

Out of curiosity were there changes in the frame rate with the previous version of this?

tcappellari-bdai:
@amessing-bdai katie had noticed yesterday the frame rate when publishing both was really slow so I made this draft and then katie went to test on a previous hash and different robots and still was seeing the same frame rate. Didnt realize it was that slow (potential bug somewhere?)

khughes-bdai:
Yeah, I was seeing a slow frame rate over ethernet yesterday on the hash in bdai yesterday (before the image compression PR went in), like 5Hz. I think it might have been an issue with my specific laptop because Jenny ran a quick test on the workstation in the spot lab with the same hash and was getting 15Hz.

amessing-bdai:
Hmm. Ok

bhung-bdai (approved):

PR #375, opened 2024-05-06
Prevent from duplicated camera_info msgs

berndgassmann:
## Change Overview

By sending both, the compressed image and the image within same function.

## Testing Done

unit test run
image transport rotation nodes used for rotating the frontcameras for Rviz don't complain (too often) on mismatches between image and camera_info topics.


tcappellari-bdai (approved):
nice catch. Thanks!

PR #388, opened 2024-05-21
[SW-1065] uncompress images using image transport

khughes-bdai:
## Change Overview

Always publish compressed images, with the option to uncompress. 

This speeds up uncompressed image publishing from ~5Hz to ~13Hz on my laptop over ethernet. Over wifi on my laptop, the speedup goes from <1Hz to ~3Hz. 

this is only when the point cloud publishing is turned off. If I enable point cloud publishing with `publish_point_clouds:=true` the compressed images still come in at 15 Hz but the regular images only at 3 (and the point clouds come in at 6 somehow). the chain goes compressed (spot_ros2) -> image (compressed_image_transport) -> point cloud (depth_image_proc)

## Testing Done

- [x] massive speed up if not publishing point clouds
- [x] successfully tested with or without `spot_name` set
- [ ] figure out how to maintain the speed while still publishing point clouds
- [ ] check speedup on a desktop machine
- [ ] check that greyscale publishing still functional

 ## questions
* There is a small slowdown when using the image republisher from image transport of about ~2Hz (I get the compressed images at 15, and the uncompressed at like 13). How much of an issue is this?
* the republisher nodes are somewhat slow to shut down after `CTRL+C`, how much of an issue is this?

### DIFF
--- a/spot_driver/include/spot_driver/api/default_image_client.hpp
+++ b/spot_driver/include/spot_driver/api/default_image_client.hpp
@@ -21,7 +21,6 @@ class DefaultImageClient : public ImageClientInterface {
                      const std::string& robot_name);
 
   [[nodiscard]] tl::expected<GetImagesResult, std::string> getImages(::bosdyn::api::GetImageRequest request,
-                                                                     bool uncompress_images,
                                                                      bool publish_compressed_images) override;
 
  private:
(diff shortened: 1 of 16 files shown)


khughes-bdai:
* **#388** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/388?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

khughes-bdai:
Closing this because I can't figure out how to maintain the speed if point clouds are published which is important. With the custom dds profile, image publishing should be coming in faster, so this is less of a priority

PR #411, opened 2024-06-14
[N/A] Update dockerfile for 4.0.2

khughes-bdai:
## Change Overview

Forgot to update this for the main change to 4.0.2. This makes it so that our CI tests use the new package versions.

## Testing Done

- [x] If CI succeeds, this is OK


mhidalgo-bdai (approved):
Looks about right.

PR #415, opened 2024-06-17
Added dockerfile for running the driver

ryan-roche:
## Change Overview

I had spoken with @khughes-bdai on a previous Issue's discussion about creating a dockerfile that can be used to run the driver in production (whether that be on a computer or pushing it to Spot CORE I/O). This is separate from the Dockerfile found in the .devcontainer directory, as that is for creating a containerized development environment for something like GitHub Codespaces.

This PR adds a dockerfile to the root of the repository that can be used to run driver within a docker container. It uses the `osrf/ros-noetic-desktop-full` image as a base, and automatically downloads and installs the driver code from this repository.

I've pushed the built image to my docker hub account for convenience. It can be found at `ryroche/spot_ros2`

## Testing Done

This PR is a bit abnormal, as it does not modify any of the code. Thus I was unable to use traditional unit testing to ensure that the additions of this PR work.

- The docker image builds successfully
- Running the container, I was able to connect the driver to my robotics lab's spot robot and call the trigger services provided by the driver (`power_on`, `stand`, `sit`, etc.)
- Graphical tools such as `rviz2` work, you just need to enable your container to connect to your X11 server, whether that be by generating/exporting XAuth data to your container or by disabling authentication by running `xauth +x`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
     python3-colcon-common-extensions \
     python3-colcon-mixin \
     python3-rosdep \
-    libpython3-dev \
-    ros-humble-tl-expected && \
+    libpython3-dev && \
     rm -rf /var/lib/apt/lists/*
 
 # Create ROS workspace
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ WORKDIR /ros_ws/src
 
 # Clone driver code
 RUN git clone https://github.com/bdaiinstitute/spot_ros2.git .
+RUN git submodule update --init --recursive
 
 # Run install script
 RUN /ros_ws/src/install_spot_ros2.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,4 +29,4 @@ RUN /ros_ws/src/install_spot_ros2.sh
 # Build packages with Colcon
 WORKDIR /ros_ws/
 RUN . /opt/ros/humble/setup.sh && \
-    colcon build --symlink-install --packages-ignore proto2ros_tests
+    colcon build --symlink-install --packages-ignore proto2ros proto2ros_tests
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
     apt-get install -yq --no-install-recommends \
     lcov \
     curl \
+    wget \
     python3-pip \
     python-is-python3 \
     python3-argcomplete \
@@ -17,9 +18,20 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
     ros-humble-tl-expected && \
     rm -rf /var/lib/apt/lists/*
 
-WORKDIR /repo
+# Create ROS workspace
+WORKDIR /ros_ws/src
 
-COPY . .
+# Clone driver code
+RUN git clone https://github.com/bdaiinstitute/spot_ros2.git .
 
+# Change gitmodule links to HTTPS to anonymously clone them
 RUN sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
 RUN git submodule update --init --recursive
+
+# Run install script
+RUN /ros_ws/src/install_spot_ros2.sh
+
+# Build packages with Colcon
+WORKDIR /ros_ws/
+RUN . /opt/ros/humble/setup.sh && \
+    colcon build --symlink-install --packages-ignore proto2ros_tests
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,6 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
     python3-colcon-mixin \
     python3-rosdep \
     libpython3-dev \
-    python3-vcstool \
     ros-humble-tl-expected && \
     rm -rf /var/lib/apt/lists/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,10 +24,6 @@ WORKDIR /ros_ws/src
 # Clone driver code
 RUN git clone https://github.com/bdaiinstitute/spot_ros2.git .
 
-# Change gitmodule links to HTTPS to anonymously clone them
-RUN sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
-RUN git submodule update --init --recursive
-
 # Run install script
 RUN /ros_ws/src/install_spot_ros2.sh
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM osrf/ros:humble-desktop-full
+
+# Install dependencies (taken from devcontainer dockerfile)
+RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
+    apt-get update -q && \
+    apt-get install -yq --no-install-recommends \
+    lcov \
+    curl \
+    python3-pip \
+    python-is-python3 \
+    python3-argcomplete \
+    python3-colcon-common-extensions \
+    python3-colcon-mixin \
+    python3-rosdep \
+    libpython3-dev \
+    python3-vcstool \
+    ros-humble-tl-expected && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /repo
+
+COPY . .
+
+RUN sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
+RUN git submodule update --init --recursive
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,6 @@ FROM osrf/ros:humble-desktop-full
 RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
     apt-get update -q && \
     apt-get install -yq --no-install-recommends \
-    lcov \
     curl \
     wget \
     python3-pip \
--- a/README.md
+++ b/README.md
@@ -57,6 +57,22 @@ source install/local_setup.bash
 
 We suggest ignoring the `proto2ros_tests` package in the build as it is not necessary for running the driver. If you choose to build it, you will see a number of error messages from testing the failure paths. 
 
+### Alternative - Docker Image
+
+Alternatively, a Dockerfile is available that prepares a ready-to-run ROS2 Humble install with the Spot driver installed.
+
+No special precautions or actions need to be taken to build the image. Just clone the repository and run `docker build` in the root of the repo to build the image.
+
+No special actions need to be taken to run the image either. However, depending on what you intend to _do_ with the driver in your project, the following flags may be useful:
+
+| Flag     | Use             |
+| -------- | --------------- |
+| `--runtime nvidia` + `--gpus all`  | Use the [NVIDIA Container Runtime](https://developer.nvidia.com/container-runtime) to run the container with GPU acceleration |
+| `-e DISPLAY`  | Bind your display to the container in order to run GUI apps. Note that you will need to allow the Docker container to connect to your X11 server, which can be done in a number of ways ranging from disabling X11 authentication entirely, or by allowing the Docker daemon specifically to access your display server.  |
+| `--network host` | Use the host network directly. May help resolve issues connecting to Spot Wifi |
+
+The image does not have the `proto2ros_tests` package built. You'll need to build it yourself inside the container if you need to use it.
+
 # Spot ROS 2 Driver
 
 The Spot driver contains all of the necessary topics, services, and actions for controlling Spot over ROS 2. To launch the driver, run:


khughes-bdai:
I am unable to build the repo when using this dockerfile -- I think it's missing the Boston dynamics package dependencies (bosdyn python packges, the bosdyn cpp sdk, and the bosdyn_msgs packages). How were you able to run this on the robot?

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    python3-vcstool \
+    ros-humble-tl-expected && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /repo
+
+COPY . .
+
+RUN sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
+RUN git submodule update --init --recursive
```
khughes-bdai, line 25: I think you will need to carry these lines over from the .devcontainer dockerfile to install the boston dynamics related requirements 

```suggestion
RUN git submodule update --init --recursive

# Install bosdyn_msgs package
RUN curl -sL https://github.com/bdaiinstitute/bosdyn_msgs/releases/download/4.0.2/ros-humble-bosdyn_msgs_4.0.2-jammy_amd64.run --output /tmp/ros-humble-bosdyn_msgs_4.0.2-jammy_amd64.run --silent \
  && chmod +x /tmp/ros-humble-bosdyn_msgs_4.0.2-jammy_amd64.run \
  && ((yes || true) | /tmp/ros-humble-bosdyn_msgs_4.0.2-jammy_amd64.run) \
  && rm /tmp/ros-humble-bosdyn_msgs_4.0.2-jammy_amd64.run

# Install spot_cpp_sdk package
RUN curl -sL https://github.com/bdaiinstitute/spot-cpp-sdk/releases/download/v4.0.2/spot-cpp-sdk_4.0.2_amd64.deb --output /tmp/spot-cpp-sdk_4.0.2_amd64.deb --silent \
  && dpkg -i /tmp/spot-cpp-sdk_4.0.2_amd64.deb \
  && rm /tmp/spot-cpp-sdk_4.0.2_amd64.deb

# Install bosdyn_msgs missing dependencies
RUN python -m pip install --no-cache-dir --upgrade pip==22.3.1 \
    && pip install --root-user-action=ignore --no-cache-dir --default-timeout=900 \
    numpy==1.24.1 \
    pytest-cov==4.1.0 \
    pytest-xdist==3.5.0 \
    bosdyn-api==4.0.2 \
    bosdyn-core==4.0.2 \
    bosdyn-client==4.0.2 \
    bosdyn-mission==4.0.2 \
    bosdyn-choreography-client==4.0.2 \
    pip cache purge

# Install spot_wrapper requirements
RUN --mount=type=bind,source=spot_wrapper,target=/tmp/context/spot_wrapper cd /tmp/context && \
    pip install --root-user-action=ignore --no-cache-dir --default-timeout=900 -r spot_wrapper/requirements.txt && \
    pip cache purge
```

ryan-roche: Interesting. I realized that I copied the repository code into the root of the ros workspace rather than the source directory, so I need to update the dockerfile regardless. I'll see what's going on.

I was able to run the dockerfile as-is and connect to my robot with no other steps. I'm as confused as you are as to how...

ryan-roche: As for the bosdyn python packages, those get taken care of by the `install_spot_ros2.sh` script when it installs the requirements for `spot_wrapper`

ryan-roche:
It might be that I had some stuff in my local repo that was getting copied into the container. I'll try and refactor it so that it isn't reliant on any local files.

Dockerfile
ryan-roche: No worries.

This dockerfile is just meant to provide a turn-key docker image for the driver that can be used as a base for projects using it in workflows like my robotics lab's where we use docker containers to manage the switching between different ROS installs (rather than switching operating systems whenever we need to jump between major ROS versions)

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM osrf/ros:humble-desktop-full
+
+# Install dependencies (taken from devcontainer dockerfile)
+RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
```
khughes-bdai, line 4: Can you add `wget` to this? This crashes my run of `install_spot_ros2.sh`:
```
./install_spot_ros2.sh: line 35: wget: command not found
```

khughes-bdai: (although I guess this should really be in the `install_spot_ros2` script)

ryan-roche: I believe I did. Must have forgotten to commit before pushing 😅

ryan-roche:
New commit- added `wget` to Dockerfile and removed `COPY` instruction- everything is done from scratch in the builder

khughes-bdai (approved):
Thanks for the PR! I was able to successfully use the Dockerfile to start the driver. 
Would you mind adding a short section to the readme that explains how to start the docker container? After that happy to merge it in

ryan-roche:
Sure- where in the readme would you like me to add the section for the Docker container?

khughes-bdai:
Maybe right before the help section?

ryan-roche:
I added the instructions for the Dockerfile inside the README. I decided it would make more sense to add it as a subsection to the Installation section

ryan-roche:
Another thought- having an image on a container registry like Docker Hub would be nice too.

There aren't any mechanisms to include user-specific code when the container is built (since that can just be done with a bind-mount when they _run_ the container), thereby making the process of having individual users build the image themselves rather redundant. As such, it would make more sense to just have an image available to be pulled and run.

Since this is a public repository, you can have GitHub Actions automatically build and push an image to Docker Hub at no cost.

I considered putting a link to the image on _my_ Docker Hub account in the README for convenience, but it feels weird having just some random guy's Image in the README for a semi-official repo (I know that Boston Dynamics AI Institute is somewhat separated from Boston Dynamics "proper", but it still feel official enough that I didn't want to just "hijack" it with my own Image link).

I can write an action YAML that would automatically build and push the image to Docker Hub whenever the main branch is updated (or on manual dispatch, which just gives you a button you can click in the actions menu to test it out manually)- all you'd need to do is add a Docker Hub token to the repo secrets, which would only be visible to the GitHub Actions Runner and the repository admins.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+# Create ROS workspace
+WORKDIR /ros_ws/src
+
+# Clone driver code
+RUN git clone https://github.com/bdaiinstitute/spot_ros2.git .
+
+# Change gitmodule links to HTTPS to anonymously clone them
+RUN sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
```
mhidalgo-bdai, line 28: @ryan-roche meta: which `ssh` URLs? I see none on [`main`](https://github.com/bdaiinstitute/spot_ros2/blob/main/.gitmodules).

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN git submodule update --init --recursive
+
+# Run install script
+RUN /ros_ws/src/install_spot_ros2.sh
+
+# Build packages with Colcon
+WORKDIR /ros_ws/
+RUN . /opt/ros/humble/setup.sh && \
+    colcon build --symlink-install --packages-ignore proto2ros_tests
```
mhidalgo-bdai, line 37: @ryan-roche nit: feel free to ignore all `proto2ros*` packages.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    wget \
+    python3-pip \
+    python-is-python3 \
+    python3-argcomplete \
+    python3-colcon-common-extensions \
+    python3-colcon-mixin \
+    python3-rosdep \
+    libpython3-dev \
+    python3-vcstool \
+    ros-humble-tl-expected && \
```
mhidalgo-bdai, line 18: @ryan-roche nit: this debian is already listed within `install_spot_ros2.sh`.

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM osrf/ros:humble-desktop-full
+
+# Install dependencies (taken from devcontainer dockerfile)
+RUN DEBIAN_FRONTEND=noninteractive apt-get update -q && \
+    apt-get update -q && \
+    apt-get install -yq --no-install-recommends \
+    lcov \
```
mhidalgo-bdai, line 7: @ryan-roche if this is a containerized Spot ROS 2 driver, it needs no test coverage tooling.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    libpython3-dev \
+    python3-vcstool \
+    ros-humble-tl-expected && \
+    rm -rf /var/lib/apt/lists/*
+
+# Create ROS workspace
+WORKDIR /ros_ws/src
+
+# Clone driver code
+RUN git clone https://github.com/bdaiinstitute/spot_ros2.git .
```
mhidalgo-bdai, line 25: @ryan-roche meta: if this image is to be built in some CD job, why not just assume the repository will be in context?

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    curl \
+    wget \
+    python3-pip \
+    python-is-python3 \
+    python3-argcomplete \
+    python3-colcon-common-extensions \
+    python3-colcon-mixin \
+    python3-rosdep \
+    libpython3-dev \
+    python3-vcstool \
```
mhidalgo-bdai, line 17: @ryan-roche I think we don't need no `vcstool`, dependencies are submodules.

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM osrf/ros:humble-desktop-full
```
mhidalgo-bdai, line 1: @ryan-roche meta: consider using a multi-stage Dockerfile to only carry over what's strictly necessary to run the driver (and leave behind all development tooling like `colcon`).

Dockerfile
mhidalgo-bdai: Also, consider using a `--merge-install` using a standard Linux FHS location (like `/opt` ) as `--install-base`. That'll help the multi-stage Dockerfile.

mhidalgo-bdai (commented):
> having an image on a container registry like Docker Hub would be nice too.

That makes sense to me. We are already pushing CI containers to our package registry. It would be straightforward to do the same for this.

Dockerfile
ryan-roche: Could you explain a bit more what exactly that would entail? I'm not too familiar with multi-stage Dockerfiles, but they look incredibly promising.

Dockerfile
ryan-roche: I will remove the install and verify that it still works

Dockerfile
ryan-roche: To be honest, I wasn't sure if you would be ok with the idea of adding an additional CD workflow, so I wrote the Dockerfile to be able to run independently.

If you're ok with adding a CD job for it, I can refactor it to use the GitHub Actions context.

Dockerfile
ryan-roche: Would you be able to detail how I would implement `--merge-install` for a multi-stage dockerfile? This driver is my first time using ROS2 and its associated tools.

Dockerfile
ryan-roche: I think I had this confused for the ROS1 driver repository...

Resolved by 661e364

Dockerfile
ryan-roche: I'm not seeing these packages within the referenced file. Are they covered by a meta package? If so, are all four lines not needed, then?

Dockerfile
ryan-roche: > @ryan-roche nit: feel free to ignore all `proto2ros*` packages.

Implemented ced6d8e

Dockerfile
ryan-roche: My idea for this Dockerfile was that it would be used as a base for projects that use the Spot driver, hence using the `desktop-full` image for ROS2 Humble. The image would include the full ROS Humble install and the precompiled driver, so all they need to do to get going with their project is mount their code into the container and build _their_ packages.

I would argue that there wouldn't be much of a reason to use the driver just on its own when things like the raw Python SDK and Tablet Controller exist- nearly everyone using the ROS driver is using it because they're working on a ROS project of their own that adds additional nodes that work with the driver, and if everyone using this image would have to manually extend it by installing the dev tools that we intentionally omitted, we might as well just leave them in the original image.

I suppose we could also make separate tags of the image that include different amounts of additional tools? One that includes the full ROS Humble install with tools like `colcon`, and another that, as you said, only contains what is strictly necessary to run the driver?

Dockerfile
mhidalgo-bdai: > Would you be able to detail how I would implement --merge-install for a multi-stage dockerfile? This driver is my first time using ROS2 and its associated tools.

Use `--merge-install` instead of `--symlink-install`, set `--install-base` to some system level path like `/opt/application`, then `COPY` over on a second stage (https://docs.docker.com/build/building/multi-stage may be informative).

Dockerfile
mhidalgo-bdai: I see. Let's see how the rest feels about a CD job then.

Dockerfile
mhidalgo-bdai: > My idea for this Dockerfile was that it would be used as a base for projects that use the Spot driver, hence using the desktop-full image for ROS2 Humble. The image would include the full ROS Humble install and the precompiled driver, so all they need to do to get going with their project is mount their code into the container and build their packages.

I see. That sounds very different from what this PR description suggests:

> ...creating a dockerfile that can be used to run the driver in **production**...

I'll re-review considering that what you are actually proposing is distributing `spot_ros2` in a Docker image for downstream users to build on top.

mhidalgo-bdai: Nevermind my first comment. A multi-stage build is overkill for the current setup.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN git clone https://github.com/bdaiinstitute/spot_ros2.git .
+RUN git submodule update --init --recursive
+
+# Run install script
+RUN /ros_ws/src/install_spot_ros2.sh
+
+# Build packages with Colcon
+WORKDIR /ros_ws/
+RUN . /opt/ros/humble/setup.sh && \
+    colcon build --symlink-install --packages-ignore proto2ros proto2ros_tests
```
mhidalgo-bdai, line 33: @ryan-roche consider installing to a known location and ensuring the binary overlay gets sourced by updating the image entrypoint. I still think a `--merge-install` is the most appropriate.

mhidalgo-bdai: Also, how would you support a user that needs to temporarily patch `spot_ros2`?

Dockerfile
ryan-roche: I see the merits in your interpretation, and I apologize if my proposal was unclear. I had assumed that it was acceptable as-is given my previous discussions. I had used the term production primarily to emphasize that this is different from the devcontainer dockerfile.

It's your repo, so you get the final say. I just wanted to explain my line of thinking. Separate tags is probably the best plan of action- that way we can have both options.

Dockerfile
ryan-roche: Gotcha. I think GitHub always shows a minimum of four lines for code references so I misinterpreted it as you also saying to remove things like `python3-rosdep`

Dockerfile
ryan-roche: Implemented f2e5779

Dockerfile
ryan-roche: I've updated the PR description to better reflect my line of thinking. I believe that there are merits to both approaches, and that it shouldn't be too difficult to build both described images and just push them with different tags (i.e. `minimal` vs `project-base`)

Dockerfile
ryan-roche: Addressed. [d3a0c7e](https://github.com/ryan-roche/spot_ros2/commit/d3a0c7e1285e493212e4ad0cc371774bc8c5281d)

Dockerfile
ryan-roche: I'm not sure how we would be able to support that workflow in a docker container like this. Wouldn't a `--merge-install` only make that more difficult?

We could run a `--merge-install` to /opt/ros/noetic so that the driver is included in the underlay regardless of where the user puts their project inside the container, but that would bring up even more issues should they need to patch the driver themselves.

I was operating on the assumption that the user would just work in the already-created `ros_ws` directory and simply rebuild `spot_ros2` if they had to patch it. Precompiling the package was more for making the image a somewhat turn-key deployment

Dockerfile
mhidalgo-bdai: Ping @khughes-bdai @amessing-bdai.

Dockerfile
mhidalgo-bdai: Circling back. Perhaps document `/ros_ws` as the well-known workspace inside the container, so that users can bind mount `spot_ros2` or their own repositories while developing. In that case, a `--merge-install` will spare you the broken links if some `spot_ros2` sources go away in that process.

Dockerfile
khughes-bdai: To be clear, this CD job would involve pushing the image to something like Docker Hub? 
I am in favor of this as I think it will make this repo easier to use -- there will just be some logistics we will need to figure out about how to set up the account, etc. (I can reach out to people about this today)

khughes-bdai:
@ryan-roche we discussed the potential CD job some more offline, and as we are currently undergoing some changes with how our docker images are stored, we think it makes the most sense to leave this PR as-is (just the Dockerfile). We can think about adding the CD job later on after this docker change goes through. 
(@mhidalgo-bdai FYI)

khughes-bdai (approved):

PR #417, opened 2024-06-25
[SW-736] Only sit on shutdown if driver holds the estop

khughes-bdai:
## Change Overview

Currently, when you CTRL+C the driver, Spot will sit. If the estop is held by the tablet, this is annoying to deal with if you are starting and stopping the driver often, so this PR removes this feature.

If the estop is held by the driver, we still need to force sitting on shutdown, otherwise the robot will collapse to the ground if you CTRL+C.

EDIT: https://github.com/bdaiinstitute/spot_wrapper/pull/122 and https://github.com/bdaiinstitute/spot_ros2/pull/420 need to go in first so I can use the blocking version of the sit command. 

## Testing Done

- [x] started driver without an estop (i.e., estop is controlled by the tablet). Upon CTRL+C'ing the driver, the robot doesn't sit.
- [x] started driver with an estop (i.e., added `start_estop: True` to my config yaml). Robot cleanly sits down upon driver CTRL+C

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -2913,7 +2913,8 @@ class SpotROS(Node):
     def destroy_node(self) -> None:
         self.get_logger().info("Shutting down ROS driver for Spot")
         if self.spot_wrapper is not None:
-            if self.spot_wrapper.check_is_powered_on():
+            # If the driver owns the estop, we want to sit before powering off.
+            if self.spot_wrapper.check_is_powered_on() and self.start_estop.value:
                 self.spot_wrapper.sit()
             self.spot_wrapper.disconnect()
         super().destroy_node()
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -2913,11 +2913,9 @@ class SpotROS(Node):
     def destroy_node(self) -> None:
         self.get_logger().info("Shutting down ROS driver for Spot")
         if self.spot_wrapper is not None:
-            # If the driver owns the estop, we want to sit before powering off.
             if self.spot_wrapper.check_is_powered_on() and self.start_estop.value:
-                self.spot_wrapper.sit()
-                # This sleep is necessary for the robot to not collapse
-                time.sleep(3.0)
+                self.get_logger().info("Sitting down...")
+                self.spot_wrapper.sit_blocking()
             self.spot_wrapper.disconnect()
         super().destroy_node()
 
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -2916,6 +2916,8 @@ class SpotROS(Node):
             # If the driver owns the estop, we want to sit before powering off.
             if self.spot_wrapper.check_is_powered_on() and self.start_estop.value:
                 self.spot_wrapper.sit()
+                # This sleep is necessary for the robot to not collapse on the ground as sit() is non-blocking
+                time.sleep(3.0)
             self.spot_wrapper.disconnect()
         super().destroy_node()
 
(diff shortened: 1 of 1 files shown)


khughes-bdai:
* **#417** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/417?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2913,8 +2913,11 @@ def step(self) -> None:
     def destroy_node(self) -> None:
         self.get_logger().info("Shutting down ROS driver for Spot")
         if self.spot_wrapper is not None:
-            if self.spot_wrapper.check_is_powered_on():
+            # If the driver owns the estop, we want to sit before powering off.
+            if self.spot_wrapper.check_is_powered_on() and self.start_estop.value:
                 self.spot_wrapper.sit()
+                # This sleep is necessary for the robot to not collapse
+                time.sleep(3.0)
```
khughes-bdai, line 2920: Unsure what to do to better address this besides this hardcoded `sleep`. The `sit()` command returns before the robot finishes sitting, even though this is a synchronous call under the hood

amessing-bdai: The spot-sdk doesn't really have synchronous calls. They have some blocking function that basically poll on requested feedback until a certain state. Should probably do that here instead of the blind sleep.

Also in the case we do not have the estop, we should probably still send the `stop` command, otherwise if spot is executing a trajectory it will continue to do so.

jbarry-bdai: Will it?  Cuz it'll lose the lease.  We should test that.

jbarry-bdai (approved):

spot_driver/spot_driver/spot_ros2.py
khughes-bdai: @jbarry-bdai will test that out this morning

khughes-bdai: @amessing-bdai is there an example of how to do that somewhere else in this repo (or spot wrapper)?

amessing-bdai: > Will it? Cuz it'll lose the lease. We should test that.

I don't think it matters that you have lost the lease if you have already sent the trajectory. We have had this happen during extract in the wud demo.

amessing-bdai: @khughes-bdai https://github.com/boston-dynamics/spot-sdk/blob/8f781a87b8816168ec59415db853934f41568620/python/bosdyn-client/src/bosdyn/client/robot_command.py#L2064

or any of the `block_until` functions in the spot-sdk

khughes-bdai: The robot _sometimes_ sits down with the addition of the `stop` command. Additionally I am noticing lease errors if I kill the driver in the middle of an action, then start it again with this change (not consistently, but often) which makes me hesitant to add this in:
```
[spot_ros2-1]     raise Exception(err_msg)
[spot_ros2-1] Exception: Lease on "body" has state (Status.OTHER_OWNER)
[spot_ros2-1] [WARN] [1719413452.923502086] [Rex.spot_ros2]: Goal state not set, assuming aborted. Goal ID: [ 66 193 196 138 203  85  71 105 185 101 203 122  49  76 163  75]

```

amessing-bdai: Can you return the lease as we close?

khughes-bdai: Isn't that accounted for in `spot_wrapper.isconnect()`? This internally calls `spot_wrapper.releaseLease()`. Or is returning the lease different from releasing it?

amessing-bdai: Ah, that returns the lease internally. I am surprised you are getting the "other owner" error. Might be a BD support time

khughes-bdai: Put up https://github.com/bdaiinstitute/spot_wrapper/pull/122 and this will make it so that we no longer need to have the `sleep`. I can make a ticket for looking into the `stop` command and the corresponding lease issues some more, might need to address this separately

mhidalgo-bdai (approved):

PR #425, opened 2024-07-08
set minimum velocity as well when setting velocity limits

skpawar1305:
with only lower max_vel set, spot still travels at full speed
this fixes it

also, these limits do not affect cmd_vel (velocity) commands

## Change Overview

## Testing Done

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -1843,8 +1843,7 @@ class SpotROS(Node):
     def handle_max_vel(self, request: SetVelocity.Request, response: SetVelocity.Response) -> SetVelocity.Response:
         """
         Handle a max_velocity service call. This will modify the mobility params to have a limit on the maximum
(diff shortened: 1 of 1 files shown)


khughes-bdai (commented):
Thanks for the PR -- just want to make sure I understand the issue. Before, if you sent a `SetVelocity` request containing some `linear.x` and sent a command to walk the robot forwards, it would stay within the limit you set, but if you walked it backwards, it would not?

skpawar1305:
> Thanks for the PR -- just want to make sure I understand the issue. Before, if you sent a `SetVelocity` request containing some `linear.x` and sent a command to walk the robot forwards, it would stay within the limit you set, but if you walked it backwards, it would not?

for spot trajectory action yes 👍

khughes-bdai:
Great, thanks for catching that! It generally looks good to me, but I want to test it out on robot to double check that our internal code won't be affected by this, and I'm unfortunately away from the office today. I can get back to you later this week

PR #430, opened 2024-07-15
[SW-822] Spot Driver without gripper

tcappellari-bdai:
## Change Overview

Added a parameter "gripperless" to indicate if the gripper is attached to the arm or not. If the arm is present but gripper is not, it will skip making the gripper and hand camera services.

Goes with https://github.com/bdaiinstitute/spot_wrapper/pull/125

## Testing Done
Tested on a Spot with a special gripperless firmware 

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -248,6 +248,8 @@ class SpotROS(Node):
         self.declare_parameter("spot_name", "")
         self.declare_parameter("mock_enable", False)
 
+        self.declare_parameter("gripperless", False)
+
         # When we send very long trajectories to Spot, we create batches of
         # given size. If we do not batch a long trajectory, Spot will reject it.
         self.declare_parameter(self.TRAJECTORY_BATCH_SIZE_PARAM, 100)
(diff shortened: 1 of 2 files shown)


bhung-bdai (approved):
I didn't realize this wasn't already a feature??

amessing-bdai (approved):

khughes-bdai:
This is maybe a topic for another PR, but how does this work with the image publishers? Eg what will happen if the robot is gripperless but the C++ image publishers try to request the hand camera?

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -428,7 +433,7 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
         has_arm = self.mock_has_arm
         if self.spot_wrapper is not None:
             has_arm = self.spot_wrapper.has_arm()
-        if has_arm:
+        if has_arm and not self.gripperless:
             all_cameras.append("hand")
         self.declare_parameter("cameras_used", all_cameras)
```
khughes-bdai, line 438: I think this param is not used at all in this node and maybe can be completely deleted

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -861,7 +867,7 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
             self.handle_graph_nav_set_localization,
             callback_group=self.group,
         )
-        if has_arm:
+        if has_arm and self.gripperless:
```
mhidalgo-bdai, line 870: @tcappellari-bdai shouldn't this be `if has_arm and not self.gripperless`?

PR #437, opened 2024-07-26
New DDS profile for a flaky test

tcappellari-bdai:
## Change Overview

Added in a new DDS profile based off of https://github.com/bdaiinstitute/bdai/pull/6448 to fix a flaky test caused by race conditions in service discovery in FastDDS.

## Testing Done

[x] I ran the pytests locally a few times and saw the flaky test pass but unsure how to robustly test it since it usually passes in CI but not always


khughes-bdai (approved):
LGTM as long as CI passes

spot_driver/test/CMakeLists.txt
```diff
@@ -130,6 +130,8 @@ target_include_directories(test_kinematic_service
 )
 target_link_libraries(test_kinematic_service spot_api)
 
+ament_add_pytest_test(spot_driver_pytest ${CMAKE_CURRENT_SOURCE_DIR} TIMEOUT 900)
```
amessing-bdai, line 133: Do we want to go with this approach of 900s for all of the tests or the one below which loops over and gives individual tests targets?

amessing-bdai: To be clear I am not leaning one way or the other. Just checking

tcappellari-bdai: i think just this one since it seems to be the only (relatively) consistent flaky test

amessing-bdai: I mean do we want one target for all python tests or one target for each python test. The former has us set a timeout for all tests. The later lets use define a timeout for each test file.

tcappellari-bdai: ohhhh, ok i misunderstood 
I think one target for all tests is fine

PR #445, opened 2024-08-05
[N/A] Workaround automatic power off on rollover

mhidalgo-bdai:
## Change Overview

When the robot rolls over to battery change pose, it powers off automatically. That can result in the otherwise successful battery change pose command to be overridden. This patch is a best-effort approach to try and cope with this. It is a workaround, as there is currently no way to unequivocally differentiate a true override from an incidental one.

## Testing Done

On real hardware, downstream.

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -1904,14 +1904,16 @@ class SpotROS(Node):
         # return None to continue processing the command feedback
         return None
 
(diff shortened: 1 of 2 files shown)


khughes-bdai (commented):
Looks like `spot_ros2/spot_ros2/spot_driver/test/pytests/test_ros_interfaces.py::SpotDriverTest::test_robot_command_goal_complete` would also need to be updated with the new argument

bhung-bdai (approved):

khughes-bdai (approved):

PR #454, opened 2024-08-19
[SW-1316] ros2 control: auto determine if robot has an arm in launchfile

khughes-bdai:
## Change Overview

* when connected to a robot, the launchfile will auto determine if the robot has an arm and use that to generate the URDF that gets used by the hardware interface (same as in `spot_driver`)
    * default controller config file that gets loaded is also now auto determined from if the robot has an arm. 
    * This required some reworking of the core launchfile to use an opaque function to add this extra logic
* Allows users to specify login info for the robot in a configuration yaml file, same as spot driver.
* Introduces a dependency on `spot_driver` so that the launchfile can use this code in `spot_driver.launch.spot_launch_helpers`. This could be reworked in the future if we decide we don't want these packages so coupled.
* minor changes to launch arguments to make the user interface simpler and more flexible, and updates readme

## Testing Done

- [x] tested on robot with arm
- [x] tested on robot without arm
- [x] tested examples in mock mode with and without mock'ed arm

### DIFF
--- a/spot_ros2_control/launch/spot_ros2_control.launch.py
+++ b/spot_ros2_control/launch/spot_ros2_control.launch.py
@@ -24,7 +24,7 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
     if hardware_interface == "robot":
         config_file = LaunchConfiguration("config_file").perform(context)
         has_arm = spot_has_arm(config_file_path=config_file, spot_name="")
-        username, password, hostname, _, _ = get_login_parameters(config_file)
+        username, password, hostname = get_login_parameters(config_file)[:3]
         login_params = f" hostname:={hostname} username:={username} password:={password}"
     else:
         has_arm = mock_has_arm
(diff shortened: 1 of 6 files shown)


tcappellari-bdai (approved):

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -11,108 +12,148 @@
...
+def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
+    hardware_interface: str = LaunchConfiguration("hardware_interface").perform(context)
+    controllers_config: str = LaunchConfiguration("controllers_config").perform(context)
+    mock_has_arm: bool = IfCondition(LaunchConfiguration("mock_has_arm")).evaluate(context)
+
+    # If connected to a physical robot, query if it has an arm. Otherwise, use the value in mock_has_arm.
+    if hardware_interface == "robot":
+        config_file = LaunchConfiguration("config_file").perform(context)
+        has_arm = spot_has_arm(config_file_path=config_file, spot_name="")
+        username, password, hostname, _, _ = get_login_parameters(config_file)
```
amessing-bdai, line 27:
```suggestion
        username, password, hostname = get_login_parameters(config_file)[:2]
```

mhidalgo-bdai (approved):
LGTM

PR #455, opened 2024-08-19
[SW-1245] : Custom Gripper Calibration Result Republisher

glvov-bdai:
## Change Overview
This allows Spot to use a calibration to register the depth image to the rgb frame
in the hand

[SW-1245] This builds on capability created in https://github.com/bdaiinstitute/spot_wrapper/pull/131#issue-2473965414
(This allows Spot to automatically calibrate the hand cameras, coming up with a stereo pair correspondence for the ToF and RGB cameras.)

### On Robot Testing
Tested the calibration sequence on a few different spots, varying the calibration parameters. Conducted drop in tests with ROS 2 bags (see below), as well as just testing running the publisher alongside all other other nodes. Verified that republished depth image creates a Point Cloud is in the same frame and orientation as the default depth image.

### Replacing default publisher with drop in replacement test
Recording relevant topics
```
ROBOT_NAME=ROBOT_NAME
ros2 bag record --output drop_in_test --topics /tf /tf_static /${ROBOT_NAME}/depth/hand/image /${ROBOT_NAME}/camera/hand/camera_info /${ROBOT_NAME}/joint_states /${ROBOT_NAME}/camera/hand/image
```
Playing it back, and adding the republisher (in separate terminals)
```
ros2 bag play drop_in_test
python3 calibrated_reregistered_hand_camera_depth_publisher.py --robot_name $ROBOT_NAME --calibration_path $CALIBRATION_PATH --topic depth_registered/hand/image
```
Then, recreating the point cloud 
```
ros2 launch spot_driver point_cloud_xyzrgb.launch.py spot_name:=ROBOT_NAME camera:=hand
```
Then, the point cloud can be viewed in RViz as it would be normally (when the QoS drop down is set to BEST EFFORT instead of RELIABLE

[SW-1245]: https://theaiinstitute.atlassian.net/browse/SW-1245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,9 +143,9 @@ python3 spot_wrapper/spot_wrapper/calibrate_spot_hand_camera_cli.py --ip <IP> -u
 Then, you can run a publisher to transform the depth image into the rgb images frame with the same image
 dimensions, so that finding the 3D location of a feature found in rgb can be as easy as passing
 the image feature pixel coordinates to the registered depth image, and extracting the 3D location.
-
+For all possible command line arguments, run ```ros2 run spot_driver calibated_reregistered_hand_camera_depth_publisher.py -h```
 ```
-ros2 run spot_ros2 calibrated_reregistered_hand_camera_depth_publisher.py --tag=default --calibration_path <SAVED_CAL> --robot_name <ROBOT_NAMESPACE> --topic_name /depth_registered/hand_custom_cal/image
+ros2 run spot_driver calibrated_reregistered_hand_camera_depth_publisher.py --tag=default --calibration_path <SAVED_CAL> --robot_name <ROBOT_NAMESPACE> --topic_name /depth_registered/hand_custom_cal/image
 ```
 
 You can treat the reregistered topic, (in the above example, ```<ROBOT_NAME>/depth_registered/hand_custom_cal/image```)
--- a/README.md
+++ b/README.md
@@ -123,10 +123,36 @@ If your image publishing rate is very slow, you can try
 > export=FASTRTPS_DEFAULT_PROFILES_FILE=<path_to_file>/custom_dds_profile.xml
 > ```
 
-## Optional Automatic Hand Stereo Camera Calibration Routine
-A custom Automatic Hand Stereo Camera Calibration Routine exists in the ```spot_wrapper``` submodule, where the
-output results can be used with ROS 2 for better RGB to Depth correspondence. 
-See the readme at ```/spot_wrapper/spot_spot_wrapper/calibration/README.md``` for more information.
+## Optional Automatic Eye-in-Hand Stereo Calibration Routine for Manipulator (Arm) Payload
+An optional custom Automatic Eye-in-Hand Stereo Calibration Routine for the arm is available for use in the ```spot_wrapper``` submodule, where the
+output results can be used with ROS 2 for improved Depth to RGB correspondence for the hand cameras.
+See the readme at [```/spot_wrapper/spot_wrapper/calibration/README.md```](https://github.com/bdaiinstitute/spot_wrapper/tree/main/spot_wrapper/calibration/README.md) for 
+target setup and relevant information.
+
+First, collect a calibration with ```spot_wrapper/spot_wrapper/calibrate_spot_hand_camera_cli.py```.
+Make sure to use the default ```--stereo_pairs``` configuration, and the default tag configuration (```--tag default```).
+
+For the robot and target setup described in [```/spot_wrapper/spot_wrapper/calibration/README.md```](https://github.com/bdaiinstitute/spot_wrapper/tree/main/spot_wrapper/calibration/README.md), the default viewpoint ranges should suffice.
+
+```
+python3 spot_wrapper/spot_wrapper/calibrate_spot_hand_camera_cli.py --ip <IP> -u user -pw <SECRET> --data_path ~/my_collection/ \
+--save_data True --result_path ~/my_collection/calibrated.yaml --photo_utilization_ratio 1 --stereo_pairs "[(1,0)]" \
+--spot_rgb_photo_width=640 --spot_rgb_photo_height=480 --tag default
+```
+
+Then, you can run a publisher to transform the depth image into the rgb images frame with the same image
+dimensions, so that finding the 3D location of a feature found in rgb can be as easy as passing
+the image feature pixel coordinates to the registered depth image, and extracting the 3D location.
+
+```
+ros2 run spot_ros2 calibrated_reregistered_hand_camera_depth_publisher.py --tag=default --calibration_path <SAVED_CAL> --robot_name <ROBOT_NAMESPACE> --topic_name /depth_registered/hand_custom_cal/image
+```
+
+You can treat the reregistered topic, (in the above example, ```<ROBOT_NAME>/depth_registered/hand_custom_cal/image```)
+as a drop in replacement by the registered image published by the default spot driver
+(```<ROBOT_NAME>/depth_registered/hand/image```). The registered depth can be easily used in tools 
+like downstream, like Open3d, (see [creating RGBD Images](https://www.open3d.org/docs/release/python_api/open3d.geometry.RGBDImage.html) and [creating color point clouds from RGBD Images](https://www.open3d.org/docs/release/python_api/open3d.geometry.PointCloud.html#open3d.geometry.PointCloud.create_from_rgbd_image)), due to matching image dimensions and registration
+to a shared frame.
 
 ## Spot CAM
 Due to known issues with the Spot CAM, it is disabled by default. To enable publishing and usage over the driver, add the following command in your configuration YAML file:
--- a/README.md
+++ b/README.md
@@ -123,6 +123,11 @@ If your image publishing rate is very slow, you can try
 > export=FASTRTPS_DEFAULT_PROFILES_FILE=<path_to_file>/custom_dds_profile.xml
 > ```
 
+## Optional Automatic Hand Stereo Camera Calibration Routine
+A custom Automatic Hand Stereo Camera Calibration Routine exists in the ```spot_wrapper``` submodule, where the
+output results can be used with ROS 2 for better RGB to Depth correspondence. 
+See the readme at ```/spot_wrapper/spot_spot_wrapper/calibration/README.md``` for more information.
+
 ## Spot CAM
 Due to known issues with the Spot CAM, it is disabled by default. To enable publishing and usage over the driver, add the following command in your configuration YAML file:
     `initialize_spot_cam: True`
--- a/spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
+++ b/spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
@@ -16,7 +16,7 @@ This script assumes you have a calibration generated with /spot_wrapper/spot_wra
 import argparse
 from typing import Optional
 
-import bdai_ros2_wrapper.scope as ros_scope
+import bdai_ros2_wrappers.scope as ros_scope
 import bdai_ros2_wrappers.process as ros_process
 import cv2
 import numpy as np
--- a/spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
+++ b/spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
@@ -223,8 +223,12 @@ def cli() -> argparse.ArgumentParser:
         help="A calibration saved with the multi-stereo charuco calibration utility",
     )
     parser.add_argument(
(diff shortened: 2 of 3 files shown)


glvov-bdai:
* **#455** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/455?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @glvov-bdai and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
```diff
@@ -0,0 +1,233 @@
...
+class CalibratedReRegisteredHandCameraDepthPublisher:
+    def __init__(
+        self,
+        calibration_path: Optional[str] = None,
+        tag: str = "default",
+        robot_name: str = "spot",
+        topic_name: str = "depth_registered/hand_custom_cal/image",
+        undistort: bool = False,
+    ):
+        self.node = ros_process.node()
```
amessing-bdai, line 38:
```suggestion
        self.node = ros_scope.ensure_node()
```

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
```diff
@@ -0,0 +1,233 @@
...
+import cv2
+import numpy as np
+import open3d as o3d
+import yaml
+from bdai_ros2_wrappers.context import wait_for_shutdown
+from cv_bridge import CvBridge
+from sensor_msgs.msg import Image
+
+
+class CalibratedReRegisteredHandCameraDepthPublisher:
```
amessing-bdai, line 29: Should this be connected to the launch file?

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+
+        reregistered_depth_topic = f"/{self.calibration['robot_name']}/{topic_name}"
+        self.node.get_logger().info(f"Creating reregistered depth publisher to {reregistered_depth_topic}")
+        self.reregistered_depth_img_pub = self.node.create_publisher(Image, reregistered_depth_topic, 10)
+
+    def republish_registered_depth_callback(self, msg: Image) -> None:
+        if self.calibration["rgb_image_dim"][0] != 0:
+            raw_depth_image = self.cv_bridge.imgmsg_to_cv2(msg, desired_encoding="passthrough")
+            pointcloud_pcd = self.depth_img_to_pointcloud(raw_depth_image, undistort=self.calibration["undistort"])
```
amessing-bdai, line 82: Should we also publish this point cloud?

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
glvov-bdai: I'm not sure on that front. 

Since it requires a calibration config file, I was hesitant to include it off of the back. I can make a default calibration so it can be included in the launch file, but then I suspect that people might have the overhead of using it without actually calibrating.

Open to suggestions on what's best here!

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
```diff
@@ -0,0 +1,233 @@
...
+            if param_name in ["camera_matrix_depth", "camera_matrix_rgb", "depth_t_rgb_R"]:
+                self.calibration[param_name] = np.array(self.calibration[param_name]).reshape((3, 3))
+            elif "dim" not in param_name and "name" not in param_name and "undistort" not in param_name:
+                self.calibration[param_name] = np.array(self.calibration[param_name])
+
+        self.cv_bridge = CvBridge()
+
+        raw_depth_topic = f"/{self.calibration['robot_name']}/depth/hand/image"
+        self.node.get_logger().info(f"Creating subscriber to raw depth at {raw_depth_topic}")
+        self.raw_depth_img_sub = self.node.create_subscription(
```
amessing-bdai, line 71: I would create the subscriber last to prevent a possible race condition where you receive an image before you create the  publisher

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
glvov-bdai: The point cloud can easily be composed independent of this republisher in the same manner that they are for the default hand. We could ROS launch the depth_image_proc xyzrgb node to do this without having to implement the projection ourselves.

I'm also down to just publish the point cloud here, but I think using depth_image_proc for this as well makes sense.

I didn't include this off of the bat for a similar reasoning as the launch file(https://github.com/bdaiinstitute/spot_ros2/pull/455#discussion_r1722328087). If we want to include the republisher in the launch, we can also put the depth_image_proc node in as well

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
```diff
@@ -0,0 +1,233 @@
...
+        depth_img = depth_img.astype(np.float32)
+
+        if undistort:
+            camera_matrix = self.calibration["camera_matrix_depth"]
+            dist_coeffs = np.array(self.calibration["dist_coeffs_depth"])
+            h, w = depth_img.shape[:2]
+            new_camera_matrix, _ = cv2.getOptimalNewCameraMatrix(camera_matrix, dist_coeffs, (w, h), 1, (w, h))
+            map1, map2 = cv2.initUndistortRectifyMap(
+                camera_matrix, dist_coeffs, None, new_camera_matrix, (w, h), cv2.CV_32FC1
+            )
```
amessing-bdai, line 110: Does this all need to be recreated each time or can this be done in the constructor?

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
amessing-bdai: We could use a flag in the launch file, so it isn't on by default but you can start it without needing a separate shell

@tcappellari-bdai @khughes-bdai Thoughts?

README.md
```diff
@@ -123,6 +123,11 @@ If your image publishing rate is very slow, you can try
 > export=FASTRTPS_DEFAULT_PROFILES_FILE=<path_to_file>/custom_dds_profile.xml
 > ```
 
+## Optional Automatic Hand Stereo Camera Calibration Routine
+A custom Automatic Hand Stereo Camera Calibration Routine exists in the ```spot_wrapper``` submodule, where the
+output results can be used with ROS 2 for better RGB to Depth correspondence. 
+See the readme at ```/spot_wrapper/spot_spot_wrapper/calibration/README.md``` for more information.
```
khughes-bdai, line 129: Not super important but this link won't work on Github  -- can we replace with the URL once the spot_wrapper PR gets merged?

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
khughes-bdai: Agree with using depth_image_proc down the line, would probably also be faster as it's using C++

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
khughes-bdai: Yeah, I think a disabled by default option in the launchfile would be useful. Maybe if the user enables it on launch, we enforce having to also pass in the camera calibration file too (and don't launch it if the calibration file isn't included). 
Might make more sense to handle that in another PR though to keep this one from becoming too big

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
amessing-bdai: I guess I was just thinking if we already generate the point cloud here then why start another process to generate the point cloud

khughes-bdai: @glvov-bdai Is there anything else that would need to be modified with `point_cloud_pcd` in order to publish it? Besides converting it to ROS and giving it a stamp, etc. 
If there's no additional processing we would need to do, I agree that it makes sense to just publish it here. (I was assuming above that there might be other steps but that could be wrong)

glvov-bdai: @khughes-bdai Before the point cloud can be used downstream, it must be colorized as well. If we want to colorize the pointcloud, we need to also subscribe to the RGB image stream (we only subscribe to raw depth here)

It may make sense to do so as we could also undistort the rgb image here based off of the loaded custom calibration intrinsic. 

However, this would require adding a time synchronizer for the RGB and depth images. I agree with @amessing-bdai that it makes sense to reduce the number of processes started but it may also increase the complexity of the publisher.

I also wonder if maybe spawning another process to generate the clouds could actually boost performance, as clouds are colorized and populated for publication parallel to/independent of creating the registered depth image (and without adding rgb rectification to the republishing loop).

I am impartial to either adding it as an depth_image_proc node, publishing it directly here from open3d, or maybe not even publishing it as we already publish the depth image. Let me know what we like best ;-p

amessing-bdai: In that case, I think start another depth image proc makes sense to not add to the complexity of this node. That can also be in the future PR that adds to the launch file.

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
mpickett-bdai: +1

glvov-bdai: ```calculate_undistortion_parameters``` now calculates these parameters, called once in the constructor

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
glvov-bdai: re-ordered in latest commit

amessing-bdai (approved):

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
```diff
@@ -0,0 +1,254 @@
...
+        if self.calibration["undistort"]:
+            self.calculate_undistortion_parameters()
+
+        self.cv_bridge = CvBridge()
+
+        reregistered_depth_topic = f"/{self.calibration['robot_name']}/{topic_name}"
+        self.node.get_logger().info(f"Creating reregistered depth publisher to {reregistered_depth_topic}")
+        self.reregistered_depth_img_pub = self.node.create_publisher(Image, reregistered_depth_topic, 10)
+
+        raw_depth_topic = f"/{self.calibration['robot_name']}/depth/hand/image"
```
khughes-bdai, line 79: The driver currently supports launching without a namespace -- if that's the case, the image topics look like `/camera/hand/image` instead of `/<RobotName>/camera/hand/image`. Can you account for that here? So if a robot name is not passed in, or is an empty string, it follows this same convention?

spot_driver/spot_driver/calibrated_reregistered_hand_camera_depth_publisher.py
```diff
@@ -0,0 +1,254 @@
...
+from cv_bridge import CvBridge
+from sensor_msgs.msg import Image
+
+
+class CalibratedReRegisteredHandCameraDepthPublisher:
+    def __init__(
+        self,
+        calibration_path: Optional[str] = None,
+        tag: str = "default",
+        robot_name: str = "spot",
```
khughes-bdai, line 35: Along with my other comment, I think it would make sense for this to be either `""` or `Optional[str] = None`

khughes-bdai (approved):
LGTM! but we will need to wait to merge this for the spot_wrapper PR to go in first, and then for dependabot to update the spot_wrapper submodule in another PR here

mpickett-bdai (approved):
LGTM

tcappellari-bdai (approved):
lgtm
i believe the spot wrapper pr got merged in yesterday so this can get merged once CI passes
Good job!

PR #461, opened 2024-08-23
[SW-1322] Namespacing support for spot_ros2_control

khughes-bdai:
## Change Overview

This PR adds namespacing for the ros2 control nodes that get launched in order to mimic the setup we have in the spot_driver. There is a new launch argument `spot_name` in `spot_ros2_control.launch.py` that will apply a namespace to all of the nodes launched by this stack.  

Discovered some fun issues with how ros2 control handles namespacing with respect to the config files for the controllers that get used. The names of the nodes in the config file must match the namespace applied. So a config file that looks like:
```
controller_manager:
  ros__parameters:
    update_rate: 333  # Hz

    joint_state_broadcaster:
      type: joint_state_broadcaster/JointStateBroadcaster
...
```
Does not actually work if you have launched the controller manager in a namespace, say `/spot/controller_manager`. Your config file needs to look like
```
spot/controller_manager:
  ros__parameters:
    update_rate: 333  # Hz

    joint_state_broadcaster:
      type: joint_state_broadcaster/JointStateBroadcaster
...
```
instead. 

Putting all of these parameters in the global namespace is also not a good idea, as the config file is designed to hold many different types of controllers. (eg, you would also want to change `forward_position_controller` to `spot/forward_position_controller`, later on in this config, and this could cause issues if there were any parameters that had the same name between these two controllers). So the only solution is to generate a config file with the name appended to it at runtime, and use that. We already have done this in the spot_driver rviz launchfile [here](https://github.com/bdaiinstitute/spot_ros2/blob/main/spot_driver/launch/rviz.launch.py) so this follows that same convention about how to modify a template file and save it. Alongside this, we also have to do the same thing if we want rviz to show the robot in different namespaces too. 

- [x] tested in mock mode: `ros2 launch spot_ros2_control spot_ros2_control.launch.py hardware_interface:=mock  mock_has_arm:=false spot_name:=test`
- [x] tested on robot: `ros2 launch spot_ros2_control spot_ros2_control.launch.py config_file:=spot_ros.yaml hardware_interface:=robot spot_name:=test`

## Testing Done


amessing-bdai:
Not for this PR, but should create clean up the config files on the exit of launch?

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -14,11 +17,78 @@
...
+        spot_name (str): Name of spot. If it's the empty string, the default controller file with no namespace is used.
+        has_arm (bool): Whether or not your robot has an arm. Necessary for defining the joints that the forward
+                        position controller should use.
+    """
+
+    arm_text = "with_arm" if has_arm else "without_arm"
+    template_filename = os.path.join(
+        get_package_share_directory(THIS_PACKAGE), "config", f"spot_default_controllers_{arm_text}.yaml"
+    )
+    output_filename = os.path.join(get_package_share_directory(THIS_PACKAGE), "config", "spot_default_controllers.yaml")
```
mhidalgo-bdai, line 40: @khughes-bdai hmm, this will break for system-level installs that are only writable by root. Consider using a temporary file instead.

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -14,11 +17,78 @@
...
+
+    with open(template_filename, "r") as template_file:
+        config = yaml.safe_load(template_file)
+
+        if spot_name:
```
mhidalgo-bdai, line 45: @khughes-bdai nit: if no `spot_name` is given, we could simply copy the file without all the YAML parsing and dumping.

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -14,11 +17,78 @@
...
+            forward_position_controller_joints = config["forward_position_controller"]["ros__parameters"]["joints"]
+            config["forward_position_controller"]["ros__parameters"]["joints"] = [
+                f"{spot_name}/{joint}" for joint in forward_position_controller_joints
```
mhidalgo-bdai, line 48: @khughes-bdai meta: not for now, because I know the controller we are using is a stock controller, but this would be so much simpler if the controller understood that a `namespace` (its own namespace?) should be applied to the given joints.

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -14,11 +17,78 @@
...
+            ]
+            config[f"{spot_name}/controller_manager"] = config["controller_manager"]
```
mhidalgo-bdai, line 50: @khughes-bdai fyi: if there's only one node in your process, you can just use a wildcard `**` for its name in parameter files. It's handy when renaming.

spot_ros2_control/launch/spot_ros2_control.launch.py
mhidalgo-bdai: Same applies to `create_rviz_config`.

spot_ros2_control/launch/spot_ros2_control.launch.py
mhidalgo-bdai: Same applies to `create_rviz_config`

mhidalgo-bdai (commented):
First pass

tcappellari-bdai (approved):
lgtm

spot_ros2_control/launch/spot_ros2_control.launch.py
khughes-bdai: Added temporary files in latest commit

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -14,11 +18,82 @@
...
+            del config["controller_manager"]
+            config[f"{spot_name}/forward_position_controller"] = config["forward_position_controller"]
+            del config["forward_position_controller"]
+        with NamedTemporaryFile(suffix=".yaml", mode="w", delete=False) as out_file:
+            yaml.dump(config, out_file)
+            return out_file.name
```
khughes-bdai, line 56: Was unsure how to handle the cleanup of the temporary file -- currently they are not deleted, otherwise they would go out of scope when this function exits and be deleted before launch can actually use them 
@mhidalgo-bdai

mhidalgo-bdai: It's fine. [`launch_ros`](https://github.com/ros2/launch_ros/blob/38152b2d04c6b6f126d31726ac9e12888b2e2e1f/launch_ros/launch_ros/actions/node.py#L369C9-L377) doesn't do any cleanup either. OS will drop them on reboot.

Problem here is `launch` not exposing good mechanisms for resource management. You could use a shutdown even handler to deal with it but it's messy.

mhidalgo-bdai (approved):
Looks about right. Ready to go with green CI.

PR #468, opened 2024-09-03
Joint Passthrough node for Action Stack Integration

llee-bdai:
## Change Overview

Node for Maple Action Stack integration:
`joint_command_passthrough` is a node that listens to `joint_states` and saves all the current states, but then pass through any new `/joint_commands` received to the `forward_position_controller`

## Testing Done

Will test on robot in the morning


llee-bdai:
@tcappellari-bdai tested on hardware and it's working in terms of that

tcappellari-bdai (approved):
lgtm

PR #473, opened 2024-09-05
Parameterize joint_command_passthrough parameters

llee-bdai:
## Change Overview

Makes the topic names parameters you can pass in (as they should be)

## Testing Done
Tested on robot

### DIFF
--- a/spot_ros2_control/examples/src/joint_command_passthrough.cpp
+++ b/spot_ros2_control/examples/src/joint_command_passthrough.cpp
@@ -20,7 +20,6 @@ class JointCommandPassthrough : public rclcpp::Node {
     JointCommandPassthrough() : Node("joint_passthrough") {
 
         std::string robot_namespace = declare_parameter("robot_namespace", "Spot");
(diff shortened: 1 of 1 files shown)


tcappellari-bdai (approved):
lgtm

spot_ros2_control/examples/src/joint_command_passthrough.cpp
```diff
@@ -18,18 +18,21 @@
 class JointCommandPassthrough : public rclcpp::Node {
  public:
   JointCommandPassthrough() : Node("joint_passthrough") {
+    std::string robot_namespace = declare_parameter("robot_namespace", "Spot");
+    std::string joint_state_topic = robot_namespace + "/" + declare_parameter("joint_state_topic", "joint_states");
```
amessing-bdai, line 22: Can you default this to empty and only have the "/" if the robot_namespace is not empty? That is the default for users who don't have multiple spots.

tcappellari-bdai (approved):

PR #501, opened 2024-10-07
[MAPLE-743] Add Cartesian arm pose commands

llee-bdai:
## Change Overview

This adds a subscriber similar to "arm_joint_commands" that listens to poses over joint angles.
Goes along with `spot_wrapper` PR here https://github.com/bdaiinstitute/spot_wrapper/pull/145

## Testing Done

Tested on Robot with various poses and different ref frames

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -569,6 +569,9 @@ class SpotROS(Node):
             self.create_subscription(
                 JointState, "arm_joint_commands", self.arm_joint_cmd_callback, 100, callback_group=self.group
             )
+            self.create_subscription(
+                PoseStamped, "arm_pose_commands", self.arm_pose_cmd_callback, 100, callback_group=self.group
+            )
 
             if not self.gripperless:
                 self.create_service(
@@ -2530,7 +2533,7 @@ class SpotROS(Node):
     def body_pose_callback(self, data: Pose) -> None:
         """Callback for cmd_vel command"""
         if self.spot_wrapper is None:
-            self.get_logger().info("Mock mode, received command vel " + str(data))
+            self.get_logger().info("Mock mode, received command pose " + str(data))
             return
         q = Quaternion()
         q.x = data.orientation.x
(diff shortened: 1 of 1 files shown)


spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2574,6 +2577,27 @@ def arm_joint_cmd_callback(self, data: JointState) -> None:
 
         self.spot_wrapper.arm_joint_cmd(**arm_joint_map)
 
+    def arm_pose_cmd_callback(self, data: PoseStamped) -> None:
+        if not self.spot_wrapper:
+            self.get_logger().info(f"mock mode, received arm pose command {data}")
```
scastro-bdai, line 2582: the most ultra nitpick (for consistency)

```suggestion
            self.get_logger().info(f"Mock mode, received arm pose command {data}")

File: spot_driver/spot_driver/spot_ros2.py [hunk 2]
```diff
@@ -569,6 +569,9 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
             self.create_subscription(
                 JointState, "arm_joint_commands", self.arm_joint_cmd_callback, 100, callback_group=self.group
             )
+            self.create_subscription(
+                PoseStamped, "arm_pose_commands", self.arm_pose_cmd_callback, 100, callback_group=self.group
```
amessing-bdai, line 573: Just out of curiosity do you want a queue size of a 100?

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -2574,6 +2577,27 @@ def arm_joint_cmd_callback(self, data: JointState) -> None:
...
+            z=data.pose.position.z,
+            qx=data.pose.orientation.x,
+            qy=data.pose.orientation.y,
+            qz=data.pose.orientation.z,
+            qw=data.pose.orientation.w,
+            ref_frame=data.header.frame_id,
+            duration=self.cmd_duration,
+            ensure_power_on_and_stand=False,
+        )
+        if not result[0]:
```
scastro-bdai, line 2596: Is there ever a situation where `result` could be an empty list, or not a list? If not, ignore.

scastro-bdai (approved):

spot_driver/spot_driver/spot_ros2.py
llee-bdai: Not necessarily.... I chose 100 because I wanted to give it a "reasonable bound" in terms of how fast we can blindly stream joint angles and have it execute as fast as possible. And I chose the same number for consistency for the cartesian pose streaming, but it makes sense to lower the pose streaming as compared to the joint angles. However, I haven't vigorously tested this yet with a full model and would be open to revisiting changing this number after using it in actual applications

spot_driver/spot_driver/spot_ros2.py
llee-bdai: For how it is written `right now` no, it will always be [bool, str] (mainly because as discussed in the other PR everything is wrapped in a try except), but always the possibility the underlying function is changed without this being updated...

amessing-bdai (approved):

PR #506, opened 2024-10-16
Add a customizable frame prefix ros param and unify the default fallback in param interface

Imaniac230:
## Change Overview

Hi,

I have updated the parameter interface with a new `frame_prefix` parameter that would allow to explicitly define the prefix for all frames. I have integrated the `spot_name` parameter into the interface as well.

I believe this would be useful especially in these cases:
* when users want to integrate the nodes into their launches, but have them organized in different namespaces (ex.:  `/spot/chassis/` and `/spot/sensors/`), which is not possible with the current default behavior
* when users want to have more fine-grained control or specific requirements for the frame names

The `spot_driver.launch.py` launch file does expose a `tf_prefix` parameter, however, it only effects the static transforms from the model URDF file. All dynamic transforms from the actual spot nodes were still broadcasting the default prefix, which was somewhat confusing when the driver was launched with a specific desired value in the `tf_prefix` launch param and was not behaving as expected.

The proposed new behavior would:
1. Always use the explicitly given frame prefix from parameters, if given.
2. If not given, then use the default `spot_name + '/'` prefix with a fallback to the node namespace if robot name is also not given (this is essentially identical to the original behavior).

The actual prefix construction logic in cpp nodes was also exported directly into a method in the parameter interface. This way all nodes can use the single definition instead of creating it independently at various different places.

### Summary
* added a `getFramePrefix()` method to parameter interface - returns an optional value from the `frame_prefix` argument
* added a `getFramePrefixWithDefaultFallback()` to the parameter interface - returns a valid prefix based on the `frame_prefix` and/or `robot_name` arguments or an empty string - this is the single place where all nodes should source the prefix from
* refactored `DefaultImageClient`, `DefaultSpotApi`, and the decompress image utils to use the more explicit `frame_prefix` argument instead of `robot_name`
* exported the local `stripPrefix()` function from object synchronizer to robot state conversions for reusability
* `SpotImagePublisher`, `Kinematic`, `ObjectSynchronizer`, `StatePublisher`, and `ImageStitcher` nodes now all use the single `getFramePrefixWithDefaultFallback()` method to get the prefix instead of each creating it independently

<s>Currently, I removed the `tf_prefix` and `spot_name` launch arguments from `spot_image_publishers.launch.py` and `spot_driver.launch.py`, but I'm guessing that they should probably be preserved for cases when a config file is not used at all?</s>

<s>I haven't updated any tests yet, nor added any meaningful new ones</s>, but I would be happy to receive any feedback on whether this whole proposal is even something that would actually be desired.

## Testing Done

I have verified the behaviors with different scenarios experimentally with a robot:
1. specifying the config parameter values
4. launching the driver: `ros2 launch spot_driver spot_driver.launch.py config_file:=spot_ros2/spot_driver/config/spot_ros_example.yaml stitch_front_images:=True`
5. observing that all frames in the tf tree are valid and published correctly, and that all expected topics publish data

This will also have to go in hand with a small addition to the `Wrapper` constructor here: https://github.com/Imaniac230/spot_wrapper/commit/eb1dc6dd7137ba570593e5219cd5eb590e7af391

### TODO
- [x] Verify that the all changes work physically with a robot as expected
- [x] Verify that all tests for the pre-commit behavior still pass as expected
  > <s>All `spot_driver` tests except the python `test_robot_state.test_joint_states` are passing OK. I am currently not able to get that one to execute correctly even from the `main` branch, before any changes.</s>

Add tests for additional scenarios:
  - [x] Default behavior without any robot name and frame prefix
  - [x] Default behavior if only robot name is given without a frame prefix
  - [x] Explicit frame prefix being correctly used if given

Maintain consistent compatibility with the original launch file arguments:
  - [x] Implement a launch helper util to handle passing the launch arguments to node parameters 
 
Finally:
- [x] Update the READMEs where relevant


Imaniac230:
Looks like there are some problems with access rights. Probably again related to this being an external PR?

Build and push docker image fails with:
```
ERROR: failed to solve: failed to push ghcr.io/bdaiinstitute/spot_ros2_jammy_humble:pr-506: unexpected status from POST request to https://ghcr.io/v2/bdaiinstitute/spot_ros2_jammy_humble/blobs/uploads/: 403 Forbidden
Error: buildx failed with: ERROR: failed to solve: failed to push ghcr.io/bdaiinstitute/spot_ros2_jammy_humble:pr-506: unexpected status from POST request to https://ghcr.io/v2/bdaiinstitute/spot_ros2_jammy_humble/blobs/uploads/: 403 Forbidden
```

Imaniac230:
Regarding the `spot_name` and `tf_prefix` parameters in launch files, I think that a simplified version of something like [this](https://github.com/ros-navigation/navigation2/blob/44a69ccde93514ed1a962b7f18cb101ec671d93a/nav2_common/nav2_common/launch/rewritten_yaml.py#L91) in launch helpers should work.

It would not have to do any of the fancy key manipulation, as the current design always just expects a wild-carded config file. It would simply check if the launch parameters are set and if so, substitute their values into the ones from the given config file. If no config file was used, it would output an empty configuration with only those corresponding parameters set. Otherwise, we either have an empty config or pass the provided config file as is.

Imaniac230:
@tcappellari-bdai I have implemented a small launch utility to work with the launch arguments. The `spot_name` usage is maintained exactly as it was originally, and the `tf_prefix` is now fully usable across the whole launch sequence as well. The launch parameters, if specified, will always take priority over parameters from a given config file.

I've updated the interface tests and rebased onto the latest `main`, so all standing todo items that I could think of should be complete, once you have some time to take a look at this.

tcappellari-bdai:
@Imaniac230 just letting you know that I haven't forgotten about this PR and I'm still working to get it to work with our current researchers' set-ups so as not to interrupt their workflow :)

Imaniac230:
@tcappellari-bdai thanks for the update. I took a second look at it now after a while, and there are some things that are not quite right. I want to test them in more detail, so I reverted it into a draft in the meantime.

P.S.: Sorry for the force-push.

Imaniac230:
Summary of changes made in https://github.com/bdaiinstitute/spot_ros2/pull/506/commits/3f2d35a6c6bf0d8bf9b51451903da5dba66ac76a and https://github.com/bdaiinstitute/spot_ros2/pull/506/commits/10ca6ae31ea5d8914a1d511ddf765a809842994b.

There were two bugs I had here:
1. I had a mistake in the launch file, where it would try to create the prefix from an empty spot name.
    * The handling of `spot_name` and `frame_prefix` params in launch files is now exported into a `get_name_and_prefix()` util in `spot_launch_helpers.py` and new unit tests are implemented for the helpers added in this PR in `test_launch_helpers.py`.
2. I had the `frame_prefix` argument in `spot_ros2`  defined with an empty default value, which meant that the `spot_name` would never actually be used, since the empty `frame_prefix` can still be used a valid option.
    * This is now correctly handled without a default value and a test was added to `test_ros_interfaces.py` to cover this.

Additional fixes:

The error message for `preferred_odom_frame` in `spot_ros2` always displayed the two valid options with the prefix:
https://github.com/bdaiinstitute/spot_ros2/blob/796bf25e9a73a56fe46f5c5bb980dc150d6ce056/spot_driver/spot_driver/spot_ros2.py#L355-L368
This could be misleading, since the non-prefixed `odom` and `vision` options are always supposed to be valid. Originally, this actually failed if the `preferred_odom_frame` was given as just `odom` or `vision`, which was fixed in previous commits.
* I extended the message to always display the full list of valid options with or without a prefix, and also added a test to check that the final `preferred_odom_frame` is always set correctly. I didn't find any actual usage of this in the python nodes, however. I also removed the two parameters `tf_name_vision_odom` and `tf_name_kinematic_odom`, since they weren't used anywhere.

The cpp nodes `state_publisher` and `object_synchronizer` did not have any validation checks for the `preferred_odom_frame` parameter:
https://github.com/bdaiinstitute/spot_ros2/blob/796bf25e9a73a56fe46f5c5bb980dc150d6ce056/spot_driver/src/robot_state/state_publisher.cpp#L36-L39
So, even though the `spot_ros2` node would fail with an error by design, these would launch without any warnings about an invalid odom frame. And, since the prefix could now be virtually any string, the parsing with the `'/'` character would also not be sufficient anymore.
* To implement checks and streamline the parsing process, I added `validatePreferredOdomFrame()` and `prependPrefix()` functions to robot state conversions. Both, `state_publisher` and `object_synchronizer`, will now check if the provided `preferred_odom_frame` is valid (given the `frame_prefix`), and if not, then fallback to a default odom frame setting and log a warning message. Unit tests for `stripPrefix()`, `prependPrefix()`, and `validatePreferredOdomFrame()` are now also implemented in `test_robot_state.cpp`.

In general, all three nodes (`spot_ros2`, `state_publisher`, `object_synchronizer`) will verify that the given `preferred_odom_frame` parameter is always one of these options: `prefix+odom`, `prefix+vision`, `odom`, and `vision`. 

One thing I'm not sure about is this section: https://github.com/bdaiinstitute/spot_ros2/blob/796bf25e9a73a56fe46f5c5bb980dc150d6ce056/spot_driver/src/conversions/robot_state.cpp#L149-L152
If I understand it correctly, these frames are coming directly from the robot messages. So it should never happen that they would be pre-pended with the prefix for ROS frames?

Once I get to the physical robot I will check that I didn't break anything with these new changes and mark this as ready. Also, I think that additional integration tests for `state_publisher` and `object_synchronizer` could be added:
1. Provide an invalid `preferred_odom_frame`
    1. Check that the warning message is correctly logged.
    2. Check that the root parent frame in the TF tree is correctly set to the default option.
2. Provide a valid `preferred_odom_frame` (different from the default option)
    1. Check that no warning mesage is logged.
    2. Check that the root parent frame in the TF tree is correctly set to the given option.

And this comment is now longer that I expected it to be.

Imaniac230:
@tcappellari-bdai just wanted to ask if this is still in consideration. Should I make any changes or update onto the latest main? I got preoccupied with other tasks, but also didn't want to make changes if there is some ongoing testing.

tcappellari-bdai:
@Imaniac230 Hi! Yes, sorry, got a little busy. Could you resolve the merge conflicts for us and then we'll test it out to confirm everything's good and hopefully merge it in. Thanks!

Imaniac230:
I've updated the branch and all conflicts are now resolved. This time I merged `main` into my branch to keep the history untouched, but let me know if you want this to be a clean rebase on top of `main` instead.

There were some minor changes:
* Refactored my added argument validation util for  `object_synchronizer` and `state_publisher` to also work with the new `tf_root` argument.
* Updated the README descriptions.
* Adjusted my additions in `get_login_parameters` to also use the literal definitions.

Otherwise, everything should have stayed the same. All tests are passing locally on my machine. I did some quick tests with the robot and all seems to work as expected.

khughes-bdai:
Hi @Imaniac230, I am having some issues getting the tests for spot driver to pass locally (i.e. `colcon test --packages-select spot_driver`). On your branch this times out for me at 15 minutes after only getting through like half the tests. On the most recent `main`, I can get through them all successfully in ~2 mins.
We have more comprehensive integration tests internally that I can also check this branch against, but want to sort out this first since it seems like something is significantly slowing things down. I'll take a look through the code to see if anything jumps out

spot_driver/include/spot_driver/conversions/robot_state.hpp
```diff
@@ -41,6 +52,59 @@ inline const std::map<const std::string, const std::string> kFriendlyJointNames
     {"arm0.wr1", "arm_wr1"},        {"arm0.f1x", "arm_f1x"},
 };
 
+static constexpr std::array<const char* const, 2> kValidOdomFrameNames{"odom", "vision"};
+
+static constexpr std::array<const char* const, 3> kValidTFRootFrameNames{"odom", "vision", "body"};
```
khughes-bdai, line 57: thanks for adding the checks for this! Might make more sense for these to live in `parameter_interface_base.hpp`, we do some similar parameter validity checking with the `cameras_used` parameter that might be a helpful reference, i.e.: 
https://github.com/bdaiinstitute/spot_ros2/blob/702c2fa967735f5cde9d7b16937aea0f408ecbbc/spot_driver/include/spot_driver/interfaces/parameter_interface_base.hpp#L67

spot_driver/include/spot_driver/conversions/robot_state.hpp
```diff
@@ -41,6 +52,59 @@ inline const std::map<const std::string, const std::string> kFriendlyJointNames
...
+  const std::string frame_with_prefix = prependPrefix(frame, prefix);
+
+  // Compare the given prefixed frame with all valid prefixed options and set false if no match is found.
+  const bool is_valid =
+      std::find_if(base_names.begin(), base_names.end(), [&prefix, &frame_with_prefix](const auto& option) -> bool {
+        return frame_with_prefix == (prefix + option);
+      }) != base_names.end();
+
+  return is_valid ? std::make_optional(frame_with_prefix) : std::nullopt;
+}
```
khughes-bdai, line 106: can you move this implementation to the .cpp file?

spot_driver/src/object_sync/object_synchronizer.cpp
```diff
@@ -329,13 +307,18 @@ ObjectSynchronizer::ObjectSynchronizer(const std::shared_ptr<WorldObjectClientIn
...
-  preferred_base_frame_with_prefix_ = preferred_base_frame_.find('/') == std::string::npos
-                                          ? frame_prefix_ + preferred_base_frame_
-                                          : preferred_base_frame_;
+  frame_prefix_ = parameter_interface_->getFramePrefixWithDefaultFallback();
+
+  const std::string tf_root = parameter_interface_->getTFRoot();
+  const std::optional<std::string> valid_tf_root =
+      validateFrameWithPrefix(tf_root, frame_prefix_, kValidTFRootFrameNames);
+  preferred_base_frame_ = stripPrefix(valid_tf_root.value_or(kValidTFRootFrameNames[0]), frame_prefix_);
+  preferred_base_frame_with_prefix_ = valid_tf_root.value_or(frame_prefix_ + kValidTFRootFrameNames[0]);
```
khughes-bdai, line 316: nit if you consolidate things in `parameter_interface_base.hpp` you can do something like:
```suggestion
  preferred_base_frame_ = stripPrefix(valid_tf_root.value_or(kDefaultTFRoot), frame_prefix_);
  preferred_base_frame_with_prefix_ = valid_tf_root.value_or(frame_prefix_ + kDefaultTFRoot);
```

https://github.com/bdaiinstitute/spot_ros2/blob/702c2fa967735f5cde9d7b16937aea0f408ecbbc/spot_driver/include/spot_driver/interfaces/parameter_interface_base.hpp#L64

Imaniac230:
> Hi @Imaniac230, I am having some issues getting the tests for spot driver to pass locally (i.e. `colcon test --packages-select spot_driver`). On your branch this times out for me at 15 minutes after only getting through like half the tests. On the most recent `main`, I can get through them all successfully in ~2 mins. We have more comprehensive integration tests internally that I can also check this branch against, but want to sort out this first since it seems like something is significantly slowing things down. I'll take a look through the code to see if anything jumps out

There is also a small addition that is needed in `spot_wrapper`.
* The new `frame_prefix` parameter was also added as an explicit (optional) argument to `spot_wrapper`.
* The `robot_name` argument in `spot_wrapper` was made optional.

I have now opened a pull request with the changes there (https://github.com/bdaiinstitute/spot_wrapper/pull/157). Sorry for that inconvenience, I didn't open that PR initially because it only made sense to me if this was going to go forward.

spot_driver/include/spot_driver/conversions/robot_state.hpp
Imaniac230: I will see if there is a clean way of resolving linker issues, when the templated implementation is moved to `.cpp`. Or there could simply be two non-templated wrapper functions (for the preferred odom frame and tf root) that would both just call the templated function internally.

spot_driver/spot_driver/launch/spot_launch_helpers.py
```diff
@@ -236,27 +243,77 @@ def get_camera_sources(config_file_path: str, has_arm: bool) -> List[str]:
...
-    username, password, hostname, port, certificate = get_login_parameters(config_file_path)
+    username, password, hostname, port, certificate, _ = get_login_parameters(config_file_path)
     gripperless = get_gripperless(get_ros_param_dict(config_file_path))
     spot_wrapper = SpotWrapper(
         username=username,
         password=password,
         hostname=hostname,
         port=port,
         cert_resource_glob=certificate,
-        robot_name=spot_name,
```
khughes-bdai, line 258: is this the only place where not setting the robot name is needed? technically whatever is passed in as the spot_name here doesn't matter at all (it could be the empty string), since we're only spinning this up temporarily to check for the presence of the arm.

spot_driver/include/spot_driver/conversions/robot_state.hpp
Imaniac230: They way it is currently implemented is that the validation still allows for the frame options to be combined with the prefix. So something like this in the config:
```
...
frame_prefix: "prefix/"
preferred_odom_frame: "prefix/vision"
tf_root: "prefix/vision"
...
```
would be valid the same way as this:
```
...
frame_prefix: "prefix/"
preferred_odom_frame: "vision"
tf_root: "vision"

...
```
It is using the given prefix together with the  `prepend` and `strip` utils for that and I didn't want to bloat the parameter interface with those things.

But this could instead be restricted to only allow specifying the base names. Then I think all of the validation could be simplified and cleanly integrated into the parameter interface. And this: https://github.com/bdaiinstitute/spot_ros2/pull/506#discussion_r1965683840 would also not be an issue anymore, since the template would never have to be used outside of the interface. The only thing that might be lost is the logger output.

Imaniac230: Commit https://github.com/bdaiinstitute/spot_ros2/pull/506/commits/6247cd96f346a93ecc114ad747121e9d149c2e3e exported all validation to the parameter interface, and is now enforcing the values to only be given as base names. Invalid values fallback to a default. I also added an rcl log warning when the given parameter is invalid.

spot_driver/include/spot_driver/conversions/robot_state.hpp
Imaniac230: With commit https://github.com/bdaiinstitute/spot_ros2/pull/506/commits/6247cd96f346a93ecc114ad747121e9d149c2e3e this is now declared and defined only in `rclcpp_parameter_interface.cpp`.

spot_driver/src/object_sync/object_synchronizer.cpp
Imaniac230: I believe that this should also be resolved with https://github.com/bdaiinstitute/spot_ros2/pull/506/commits/6247cd96f346a93ecc114ad747121e9d149c2e3e.

khughes-bdai:
Thanks for the updates! Will plan to test this branch some more after the weekend

khughes-bdai:
@Imaniac230 all the tests in this repo are passing locally for me, in ~ the same amount of time. I also just tested this branch against some of our internal integration ROS tests that use the driver. All of those are passing too, so this seems like it hasn't broken any existing user interfaces :tada: will plan for a more thorough review of the code next week. Thanks for your patience!

Imaniac230:
@khughes-bdai glad to hear that, thanks for taking the time to review this.

khughes-bdai:
Hey! So sorry to ask this again, but can you update this PR with the latest main? There was a large chunk of work recently merged in -- I think the main thing this will affect here is that the launch helpers have been moved to a new package `spot_common`. Thanks!

Imaniac230:
@khughes-bdai I updated the branch, there wasn't that much conflicts in the end. I moved the tests that I had created for the launch helpers into the new `spot_common` package. Otherwise there are no differences. I ran all tests locally and tested with the robot. All seems to work as before.

spot_common/spot_common/launch/spot_launch_helpers.py
```diff
@@ -260,3 +266,55 @@ def spot_has_arm(config_file_path: str, spot_name: str) -> bool:
...
+    from the yaml config or passed through from launch arguments. This will compose the frame prefix from the Spot name
+    if not given explicitly.
+
+    Args:
+        ros_params (dict[str, Any]): A dictionary of parameter_name: parameter_value.
+
+    Returns:
+        Tuple[str | Substitution, str | Substitution]: spot_name, tf_prefix.
+    """
+    spot_name: Union[str, Substitution] = ros_params["spot_name"] if "spot_name" in ros_params else ""
```
khughes-bdai, line 310: nit: can you use the `_SPOT_NAME` constant here? You can also define another one for `_FRAME_PREFIX` just below.

spot_common/spot_common/launch/spot_launch_helpers.py
```diff
@@ -260,3 +266,55 @@ def spot_has_arm(config_file_path: str, spot_name: str) -> bool:
...
+        config_file_path if isinstance(config_file_path, str) else config_file_path.perform(context)
+    )
+    for key, value in substitutions.items():
+        if value.perform(context):
+            config_params[key] = value
+
+    return config_params
+
+
+def get_name_and_prefix(ros_params: Dict[str, Any]) -> Tuple[Union[str, Substitution], Union[str, Substitution]]:
```
khughes-bdai, line 299: Question/nit: Is there a reason to keep the Substitution type in the return value? If it's `Tuple[str, str]` then you wouldn't need to call `perform(context)` on the results

spot_common/spot_common/launch/spot_launch_helpers.py
Imaniac230: Yeah, fixed in https://github.com/bdaiinstitute/spot_ros2/pull/506/commits/3b0043c9e38e65053dd62f02eacbb9fd35b26450. Completely forgot about those, thanks.

spot_common/spot_common/launch/spot_launch_helpers.py
Imaniac230: Yes, most places where the output is being used in launchfiles also supports both types, so I didn't want to make any constraints to strings here. The only reason it's explicitly converted to a string in `spot_image_publishers.launch.py` is because it is being used as a string here:
https://github.com/bdaiinstitute/spot_ros2/blob/b6e749709c5daf99940b68c45426efb7c61bf886/spot_driver/launch/spot_image_publishers.launch.py#L167
otherwise all other usages are agnostic so the perform would not be necessary.

khughes-bdai (approved):
LGTM! Just re-tested again against some external code and everything is still working great. Thanks for all the work on this!

tcappellari-bdai (approved):

PR #516, opened 2024-10-22
[SW-1466] ability to take joint gains in through a parameter file

khughes-bdai:
## Change Overview

1. standardizes gain names in the hardware interface to `k_q_p` and `k_qd_p`, same as spot sdk (https://dev.bostondynamics.com/docs/concepts/joint_control/readme)

2. allows taking in `k_q_p` and `k_qd_p` through a a parameter file. If you have a yaml that looks like this
```
/**:
  ros__parameters:
    k_q_p: [624.0, 936.0, 286.0,  624.0, 936.0, 286.0, 624.0, 936.0, 286.0, 624.0, 936.0, 286.0, 1020.0, 255.0, 204.0, 102.0, 102.0, 102.0, 16.0]
    k_qd_p: [5.20, 5.20, 2.04, 5.20, 5.20, 2.04, 5.20, 5.20, 2.04, 5.20, 5.20, 2.04, 10.2, 15.3, 10.2, 2.04, 2.04, 2.04, 0.32]
```
and pass it to the `spot_ros2_control.launch.py` launchfile with the arg `config_file:=<file path>`, these gains will get used in the hardware interface to stream commands.
[Because ros2 control hardware interfaces don't have "true" support for parameters, ](https://github.com/ros-controls/ros2_control/issues/347)the gains have to be passed in as a [string through the xacro](https://github.com/ros-controls/ros2_control/blob/cb91599f8f66aaf39b7485a2f7e131157f633474/hardware_interface/include/hardware_interface/hardware_info.hpp#L182), so this also adds a helper function to assist with this conversion back to a usable vector. 

edge case handling:
1. if either of these params isn't set (or no config file is specified) then the default gains are used. 
2. if either the gain lists defined in the parameter file is the wrong number of elements, the hardware interface will fall back to using the default gains, and a warning will be logged about this.

## Testing Done

* The wiggle arm example has some slightly jerky gripper motion using the default gains.
I changed the `k_qd_p` gain for the gripper joint in my parameter file from `0.32` to `1.32` and reran the example, and the gripper looked a lot smoother. 

* Checked edge case handling manually 

### Next steps

The function to translate the gains as a string into a usable vector in the hardware interface should get unit tested (SW-1148)

### DIFF
--- a/spot_description/urdf/spot.ros2_control.xacro
+++ b/spot_description/urdf/spot.ros2_control.xacro
@@ -20,10 +20,6 @@
                 <param name="min">${-effort_max}</param>
                 <param name="max">${effort_max}</param>
             </command_interface>
-            <!-- TODO at some point, this should be a part of the command interface so we can have a 
-            one-to-one mapping to the spot sdk implementation. -->
-            <!-- <command_interface name="k_q_p" />
-            <command_interface name="k_qd_p" /> -->
         </joint>
     </xacro:macro>
 
@@ -65,7 +61,7 @@
 
     </xacro:macro>
 
-    <xacro:macro name="spot_ros2_control" params="interface_type has_arm hostname username password tf_prefix">
+    <xacro:macro name="spot_ros2_control" params="interface_type has_arm hostname username password tf_prefix kp kd">
         <!-- Currently implements a simple system interface that covers all joints of the robot.
         In the future, we could make different hardware interfaces for the body, arm, etc. -->
         <ros2_control name="SpotSystem" type="system">
@@ -80,6 +76,8 @@
                 <param name="hostname">$(optenv SPOT_IP ${hostname})</param>
                 <param name="username">$(optenv BOSDYN_CLIENT_USERNAME ${username})</param>
                 <param name="password">$(optenv BOSDYN_CLIENT_PASSWORD ${password})</param>
+                <param name="kp">${kp}</param>
+                <param name="kd">${kd}</param>
             </xacro:if>
             </hardware>
             <!-- Add the legs -->
--- a/spot_description/urdf/spot.ros2_control.xacro
+++ b/spot_description/urdf/spot.ros2_control.xacro
@@ -61,7 +61,7 @@
 
     </xacro:macro>
 
-    <xacro:macro name="spot_ros2_control" params="interface_type has_arm hostname username password tf_prefix kp kd">
+    <xacro:macro name="spot_ros2_control" params="interface_type has_arm hostname username password tf_prefix k_q_p k_qd_p">
         <!-- Currently implements a simple system interface that covers all joints of the robot.
         In the future, we could make different hardware interfaces for the body, arm, etc. -->
         <ros2_control name="SpotSystem" type="system">
@@ -76,8 +76,8 @@
                 <param name="hostname">$(optenv SPOT_IP ${hostname})</param>
                 <param name="username">$(optenv BOSDYN_CLIENT_USERNAME ${username})</param>
                 <param name="password">$(optenv BOSDYN_CLIENT_PASSWORD ${password})</param>
-                <param name="kp">${kp}</param>
-                <param name="kd">${kd}</param>
+                <param name="k_q_p">${k_q_p}</param>
+                <param name="k_qd_p">${k_qd_p}</param>
             </xacro:if>
             </hardware>
             <!-- Add the legs -->
--- a/spot_description/urdf/spot.urdf.xacro
+++ b/spot_description/urdf/spot.urdf.xacro
@@ -17,8 +17,8 @@
   <xacro:arg name="hostname" default="10.0.0.3" />
   <xacro:arg name="username" default="username" />
   <xacro:arg name="password" default="password" />
-  <xacro:arg name="kp" default="" />
-  <xacro:arg name="kd" default="" />
+  <xacro:arg name="k_q_p" default="" />
+  <xacro:arg name="k_qd_p" default="" />
 
   <!-- Load Spot -->
   <xacro:load_spot
@@ -34,8 +34,8 @@
                                username="$(arg username)" 
                                password="$(arg password)" 
                                tf_prefix="$(arg tf_prefix)" 
-                               kd="$(arg kd)" 
-                               kp="$(arg kp)" />
+                               k_q_p="$(arg k_q_p)" 
+                               k_qd_p="$(arg k_qd_p)" />
   </xacro:if>
 
 </robot>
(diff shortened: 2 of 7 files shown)


khughes-bdai:
* **#516** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

spot_description/urdf/spot.urdf.xacro
```diff
@@ -14,10 +14,11 @@
   <!-- Parameters for ROS 2 control -->
   <xacro:arg name="add_ros2_control_tag" default="false" />
   <xacro:arg name="hardware_interface_type" default="mock" />
-  <xacro:arg name="tf_prefix" default="" />
```
khughes-bdai, line 17: (this is a duplicated arg, deleted)

spot_hardware_interface/src/spot_hardware_interface.cpp
```diff
@@ -65,6 +82,29 @@ hardware_interface::CallbackReturn SpotHardware::on_init(const hardware_interfac
...
+  } else {
+    RCLCPP_ERROR(rclcpp::get_logger("SpotHardware"),
+                 "Got %ld joints, expected either %d (Spot with arm) or %d (Spot without arm)!!", njoints_, kNjointsArm,
+                 kNjointsNoArm);
+    return hardware_interface::CallbackReturn::ERROR;
+  }
+
+  // Get the user-passed in k_q_p and k_q_d values.
+  const std::string k_q_p_string = info_.hardware_parameters["k_q_p"];
+  const std::string k_qd_p_string = info_.hardware_parameters["k_qd_p"];
```
mhidalgo-bdai, line 102: @khughes-bdai what if either parameter is not defined?

mhidalgo-bdai (approved):
Overall LGTM

spot_hardware_interface/src/spot_hardware_interface.cpp
khughes-bdai: If either isn't defined, these values are just the empty string. It ends up going through the rest of the conversion pipeline as an empty vector, which is then ignored in favor of the default gains since it doesn't have the right number of elements.

spot_description/urdf/spot.ros2_control.xacro
```diff
@@ -65,7 +61,7 @@
 
     </xacro:macro>
 
-    <xacro:macro name="spot_ros2_control" params="interface_type has_arm hostname username password tf_prefix">
+    <xacro:macro name="spot_ros2_control" params="interface_type has_arm hostname username password tf_prefix k_q_p k_qd_p">
```
robodreamer, line 64: nit: Just checking whether the gains should be named to `kp` and `kd` gains.

spot_hardware_interface/include/spot_hardware_interface/spot_constants.hpp
```diff
@@ -29,19 +29,19 @@ inline constexpr int kNjointsArm = 19;
...
-/// @brief Default Kd gains for robot without an arm
-inline constexpr float kDefaultKdNoArm[] = {5.20, 5.20, 2.04, 5.20, 5.20, 2.04, 5.20, 5.20, 2.04, 5.20, 5.20, 2.04};
+/// @brief Default k_qd_p gains for robot without an arm
+inline constexpr float kDefaultKqdpNoArm[] = {5.20, 5.20, 2.04, 5.20, 5.20, 2.04, 5.20, 5.20, 2.04, 5.20, 5.20, 2.04};
 
-/// @brief Default Kp gains for robot with an arm
-inline constexpr float kDefaultKpArm[] = {624.0, 936.0, 286.0,  624.0, 936.0, 286.0, 624.0, 936.0, 286.0, 624.0,
-                                          936.0, 286.0, 1020.0, 255.0, 204.0, 102.0, 102.0, 102.0, 16.0};
+/// @brief Default k_q_p gains for robot with an arm
+inline constexpr float kDefaultKqpArm[] = {624.0, 936.0, 286.0,  624.0, 936.0, 286.0, 624.0, 936.0, 286.0, 624.0,
```
robodreamer, line 40: It seems that currently the default gains for the leg joints are the same for both robots. Should we just keep one vector of the leg joint gains in that case if we are not going to change these default gains?

spot_description/urdf/spot.ros2_control.xacro
khughes-bdai: Yep, I renamed them to match spot sdk conventions: https://dev.bostondynamics.com/docs/concepts/joint_control/readme

robodreamer (approved):
Yay so glad to see this update! Looking nice, thanks Katie! Should there be a default config file included?

spot_ros2_control/README.md
```diff
@@ -24,6 +24,14 @@ You can then run the launchfile with the following command:
 ros2 launch spot_ros2_control spot_ros2_control.launch.py hardware_interface:=robot config_file:=path/to/spot_ros.yaml
 ```
 
+Joint level gains can also be specified in this same config file under the parameter names `k_q_p` and `k_qd_p`. If you do not specify these parameters, the default gains from the `spot-sdk` joint control examples are used during command streaming. The example below shows a valid configuration for a robot with an arm as each list contains 19 elements. More information on how these gains are used by Spot can be found [here](https://dev.bostondynamics.com/docs/concepts/joint_control/readme).
```
robodreamer, line 27: Should there be a default config file included? Do we show an example with the CLI argument for the config?

khughes-bdai: The CLI argument is the same as the command just above this paragraph (`config_file:=path/to/config.yaml'). I don't want there to be a default config file since a) a common use case for external users is to put their login info in the config file, which they are going to want to customize and b) there would have to be more logic in the launchfile to handle different default config files for robots with and without the arm since they have different #s of elements. If no config file is used then the defaults are picked up by the hardware interface

spot_hardware_interface/include/spot_hardware_interface/spot_constants.hpp
khughes-bdai: Yeah, I thought about this too. I think structuring it in this way makes it easier to handle the logic in the hardware interface. so instead of parsing together a list of the gains from two separate sources (i.e. a list of arm gains and a list of leg gains), I can just get all of the gains from the same variable at once.

robodreamer: I see. Maybe we could at least mention in the comment that the numerical values are identical currently.

robodreamer (approved):

mhidalgo-bdai (approved):

khughes-bdai:
### Merge activity

* **Oct 29, 9:25 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/516).

PR #531, opened 2024-11-26
[MAPLE-681] Add robot_description_package as parameter to driver launch

llee-bdai:
## Change Overview
This allows you to change the pack from where to get the robot description. This allows us to have different configurations of the robot (such as custom payloads)
Makes the (hard) assumption that whatever package you point it to has `{package}/urdf/spot.urdf.xacro` in it.

@tcappellari-bdai @khughes-bdai feel free to agree/disagree with this, but only way I could get extra collision objects in as needed on hardware

### DIFF
--- a/spot_driver/launch/spot_driver.launch.py
+++ b/spot_driver/launch/spot_driver.launch.py
@@ -23,6 +23,7 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
     spot_name = LaunchConfiguration("spot_name").perform(context)
     tf_prefix = LaunchConfiguration("tf_prefix").perform(context)
     mock_enable = IfCondition(LaunchConfiguration("mock_enable", default="False")).evaluate(context)
+    robot_description_package = LaunchConfiguration("robot_description_package").perform(context)
 
     # if config_file has been set (and is not the default empty string) and is also not a file, do not launch anything.
     config_file_path = config_file.perform(context)
@@ -35,7 +36,7 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
     else:
         has_arm = spot_has_arm(config_file_path=config_file.perform(context), spot_name=spot_name)
 
-    pkg_share = FindPackageShare("spot_description").find("spot_description")
+    robot_description_pkg_share = FindPackageShare(robot_description_package).find(robot_description_package)
 
     # Since spot_image_publisher_node is responsible for retrieving and publishing images, disable all image publishing
     # in spot_driver.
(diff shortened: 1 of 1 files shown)


khughes-bdai:
@amessing-bdai this is the alternative I thought of that uses xacro args in the main spot URDF. it should allow you to not duplicate the rest of the spot urdf internally but I don't know if this is actually a better solution 

```
...
<xacro:arg name="extras_package" default=""/>
...
<load main spot model>

<xacro:if value="$(arg extras_package)">
  <xacro:include filename="$(find extras_package)/urdf/spot_extras.urdf" />
```
then in the launchfile we could have 
```
robot_description = Command(
        [
            PathJoinSubstitution([FindExecutable(name="xacro")]),
            " ",
            PathJoinSubstitution([pkg_share, "urdf", "spot.urdf.xacro"]),
            " ",
            "arm:=",
            TextSubstitution(text=str(has_arm).lower()),
            " ",
            "tf_prefix:=",
            tf_prefix,
            " ",
            "extras_package:=",
           LaunchConfiguration(robot_description_package)
        ]
    )
```

there is also this leftover remnant from the ros1 repo https://github.com/bdaiinstitute/spot_ros2/blob/main/spot_description/urdf/empty.urdf but setting an environment variable to add extra things to your URDF seems like not a great way to do things

khughes-bdai (approved):
I'm good with this change, by default this doesn't change any behavior. I think this will also be a useful addition for driver integration with the custom gripper Spot. (and with the future push to separate out spot description and consolidate all of our urdfs in RDS, we might come up with a more streamlined way to add to the base spot urdf)

amessing-bdai:
It seems better to have addons vs duplicate urdf files to me, but I wouldn't be a user. @llee-bdai thoughts?

Consolidating in RDS has some hurdles to make happen, so I wouldn't base a decision on it.

khughes-bdai:
i think the downside of add-ons is that it's harder to expose custom arguments through the core spot URDF, so it feels less flexible

(with the custom gripper work as an example, I would want to set the xacro args `gripperless` and `custom_gripper_base_link` in order to get an add-on to the main spot URDF to work, which is not exposed through the launchfile's xacro command)

scastro-bdai (approved):
While I normally would agree that xacro args is better, I'm not entirely sure these "antlers" warrant being an officially supported argument until it becomes more of a standard thing. So I'm onboard with "different URDF" for the short term.

khughes-bdai (approved):

llee-bdai:
### Merge activity

* **Dec 2, 4:14 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/531).

Issue #527, opened 2024-11-21
Failing to establish time-sync


ryan-roche:
When running the driver using the provided launchfile, I'm getting a bunch of errors like this, and none of the cameras working through the driver.
```
[state_publisher_node-5] [ERROR] [1732218111.989386028] [state_publisher]: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[state_publisher_node-5] [ERROR] [1732218112.009387522] [state_publisher]: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_ros2-1] [INFO] [1732218112.585374652] [spot_ros2]: Found estop!
[spot_ros2-1] [INFO] [1732218112.790193254] [spot_ros2]: Driver successfully started!
[spot_image_publisher_node-7] [ERROR] [1732218113.065252920] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_image_publisher_node-7] [ERROR] [1732218115.899176041] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_image_publisher_node-7] [ERROR] [1732218120.097876132] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_image_publisher_node-7] [ERROR] [1732218127.657682809] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_image_publisher_node-7] [ERROR] [1732218131.367276749] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_image_publisher_node-7] [ERROR] [1732218133.961987812] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_image_publisher_node-7] [ERROR] [1732218136.771661132] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_image_publisher_node-7] [ERROR] [1732218139.077671862] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
[spot_image_publisher_node-7] [ERROR] [1732218141.378089807] [image_publisher]: Failed to get images: Failed to get latest clock skew: Time sync not yet established between Spot and the local system.
```

[...]

PR #544, opened 2025-01-03
Fixes TimeSync with robot

mschweig:
## Change Overview
The DefaultTimeSyncApi is missing the `WaitForSync` call from Spot SDK. As a result the timesync cannot be established, especially when connecting via WIFI.

I added the  `WaitForSync` call to the DefaultTimeSyncApi with a timeout of 5 seconds (maybe create a config param for that) before checking the timesync with `HasEstablishedTimeSync`

This results in a successful timesync even via WIFI over VPNs. 

Screenshot from 2025-01-03 17-41-07

Corresponding issue:
https://github.com/bdaiinstitute/spot_ros2/issues/527

## Testing Done

1. Launched the driver with various parameters
2. Tested the image services
3. Tested robot commands such as dock and undock

### DIFF
--- a/spot_driver/src/api/default_time_sync_api.cpp
+++ b/spot_driver/src/api/default_time_sync_api.cpp
@@ -1,8 +1,8 @@
 // Copyright (c) 2023-2024 Boston Dynamics AI Institute LLC. All rights reserved.
 
+#include <chrono>
 #include <spot_driver/api/default_time_sync_api.hpp>
 #include <tl_expected/expected.hpp>
-#include <chrono>
 
 namespace spot_ros2 {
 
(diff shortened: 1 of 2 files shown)


khughes-bdai:
Thank you for the PR! It looks like the linting CI check is failing. On your branch, could you please run `pre-commit install` then `pre-commit run --all` to fix this?

mschweig:
Hi @khughes-bdai

happy to help :) ran the linter and added the sign-off so that DCO does not complain. 

CI should be ok now.

khughes-bdai (approved):
Looks good! tested on robot and also tested running CI locally (the failing CI test currently is a known issue for external contributors).
Thanks!

mschweig:
Hi @khughes-bdai, I would like to improve it a bit more and set the timeout value as a parameter, so that its not hardcoded to the `TimeSyncApi`. Should I add it as a `ParameterInterfaceBase` parameter or do you recommend another approach? 

Thanks.

khughes-bdai:
Hi @mschweig, sounds good and yes `ParameterInterfaceBase` would be the right place to put it. Thanks!

PR #552, opened 2025-01-10
Adds timesync timeout to params

mschweig:
## Change Overview

PR https://github.com/bdaiinstitute/spot_ros2/pull/544 added the `WaitForSync` call to establish a timesync with the robot. This function has a `timeout` parameter which was hardcoded before. This PR now adds the timeout to the `ParameterInterface`

I needed to add a rclcpp node to `default_spot_api` to get the parameter values, not sure if this is how you intended the use of the `ParameterInterface`. If you like another approach I am happy to change it :)

## Testing Done

1. tested various timeouts
2. connected to a robot via WAN VPN
3. executed locomotion commands

### DIFF
--- a/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
+++ b/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
@@ -199,7 +199,9 @@ bool RclcppParameterInterface::getGripperless() const {
 }
 
 std::chrono::seconds RclcppParameterInterface::getTimeSyncTimeout() const {
-  int timeout_seconds = declareAndGetParameter<int>(node_, kParameterTimeSyncTimeout, std::chrono::duration_cast<std::chrono::seconds>(kDefaultTimeSyncTimeout).count());
+  int timeout_seconds =
+      declareAndGetParameter<int>(node_, kParameterTimeSyncTimeout,
+                                  std::chrono::duration_cast<std::chrono::seconds>(kDefaultTimeSyncTimeout).count());
   return std::chrono::seconds(timeout_seconds);
 }
 


spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
```diff
@@ -28,6 +28,7 @@ constexpr auto kParameterNamePublishDepthRegisteredImages = "publish_depth_regis
 constexpr auto kParameterPreferredOdomFrame = "preferred_odom_frame";
 constexpr auto kParameterTFRoot = "tf_root";
 constexpr auto kParameterNameGripperless = "gripperless";
+constexpr auto kParameterTimeSyncTimeout = "timeout";
```
khughes-bdai, line 31: nit: can you call this `timesync_timeout`?

spot_driver/src/api/default_spot_api.cpp
```diff
@@ -55,6 +62,10 @@ tl::expected<void, std::string> DefaultSpotApi::authenticate(const std::string&
   }
   // Start time synchronization between the robot and the client system.
   // This must be done only after a successful authentication.
+  // Create a rclcpp node and parameter interface to retrieve ROS2 parameters, such as timesync_timeout.
+  const auto node_ = std::make_shared<rclcpp::Node>(kNodeName, kNodeNameSpace);
+  auto parameter_interface = std::make_unique<RclcppParameterInterface>(node_);
+  const auto timesync_timeout = parameter_interface->getTimeSyncTimeout();
```
khughes-bdai, line 68: I don't love the idea of having to make a node to get this parameter. How about getting this parameter from something that's already a node, then pass it along as an optional arg to the constructor of `DefaultSpotApi`? (e.g. in `kinematic_node.cpp`, `object_synchronizer_node.cpp`, `state_publisher_node.cpp`, and `spot_image_publisher_node.cpp`).

khughes-bdai:
Thanks for the PR! Along with the other comments, could you please also add to the unit tests in `spot_driver/test/src/test_parameter_interface.cpp` with this new parameter?

spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
mschweig: oops, sure :)

spot_driver/src/api/default_spot_api.cpp
mschweig: Yeah, you're right, don't love this either. Also thought about this approach, but the existing nodes (kinematic, obj sync, state pub and image pub) are creating their own unique pointers of `<DefaultSpotApi>` so that every node has its own time sync established. In this case we have to make sure every DefaultSpotApi has the `timesync_timeout` properly set, so it should be a required parameter of the `DefaultSpotApi` constructor, right?

spot_driver/src/api/default_spot_api.cpp
```diff
@@ -25,6 +26,9 @@ DefaultSpotApi::DefaultSpotApi(const std::string& sdk_client_name, const std::op
   } else {
     client_sdk_ = ::bosdyn::client::CreateStandardSDK(sdk_client_name);
   }
+  if (timesync_timeout.has_value()) {
+    timesync_timeout_ = timesync_timeout.value();
+  }
```
khughes-bdai, line 31: realizing an issue with making this optional -- what to do in the case that it is does not have a value? I think it might make more sense to just enforce this as a required argument

mschweig: yeah, wanted to make sure its okay for you to enforce this parameter, before implementing it. Gonna do that :)
It will be an issue if it's optional.

spot_driver/src/images/spot_image_publisher_node.cpp
```diff
@@ -39,7 +39,8 @@ SpotImagePublisherNode::SpotImagePublisherNode(const rclcpp::NodeOptions& node_o
   auto tf_broadcaster = std::make_unique<RclcppTfBroadcasterInterface>(node);
   auto timer = std::make_unique<RclcppWallTimerInterface>(node);
 
-  auto spot_api = std::make_unique<DefaultSpotApi>(kSDKClientName, parameters->getCertificate());
+  auto timesync_timeout = parameters->getTimeSyncTimeout();
```
khughes-bdai, line 42: nit: this (and other calls to getTimeSyncTimeout in other nodes) can be const 
```suggestion
  const auto timesync_timeout = parameters->getTimeSyncTimeout();
```

spot_driver/src/api/default_time_sync_api.cpp
```diff
@@ -6,14 +6,15 @@
...
+DefaultTimeSyncApi::DefaultTimeSyncApi(std::shared_ptr<::bosdyn::client::TimeSyncThread> time_sync_thread,
+                                       int8_t timesync_timeout)
+    : time_sync_thread_{time_sync_thread}, timesync_timeout_{timesync_timeout} {}
 
 tl::expected<google::protobuf::Duration, std::string> DefaultTimeSyncApi::getClockSkew() {
   if (!time_sync_thread_) {
     return tl::make_unexpected("Time sync thread was not initialized.");
   }
-  if (!time_sync_thread_->WaitForSync(std::chrono::seconds(5))) {
+  if (!time_sync_thread_->WaitForSync(std::chrono::seconds(timesync_timeout_))) {
```
bjin-bdai, line 17: given this API uses `chrono`, I believe the units for `timesync_timeout_` should also be a `chrono` object instead of `int8_t`. The nice thing about the chrono library is that the unit conversion are handled for you by the compiler

spot_driver/src/api/default_time_sync_api.cpp
```diff
@@ -6,14 +6,15 @@
 
 namespace spot_ros2 {
 
-DefaultTimeSyncApi::DefaultTimeSyncApi(std::shared_ptr<::bosdyn::client::TimeSyncThread> time_sync_thread)
-    : time_sync_thread_{time_sync_thread} {}
+DefaultTimeSyncApi::DefaultTimeSyncApi(std::shared_ptr<::bosdyn::client::TimeSyncThread> time_sync_thread,
+                                       int8_t timesync_timeout)
```
bjin-bdai, line 10:
```suggestion
                                       std::chrono::milliseconds timesync_timeout)
```
Doesn't have to be milliseconds, but chrono instead of int8_t seems more appropriate here

spot_driver/include/spot_driver/api/default_spot_api.hpp
```diff
@@ -20,7 +20,8 @@ namespace spot_ros2 {
 class DefaultSpotApi : public SpotApi {
  public:
   explicit DefaultSpotApi(const std::string& sdk_client_name,
-                          const std::optional<std::string>& certificate = std::nullopt);
+                          const std::optional<std::string>& certificate = std::nullopt,
+                          const std::optional<int8_t> timesync_timeout = std::nullopt);
```
bjin-bdai, line 24:
```suggestion
                          const std::optional<std::chrono::milliseconds> timesync_timeout = std::nullopt);
```
chrono seems like a better library for dealing with time

bjin-bdai (commented):
not to familiar with the stack, but based on the diff it seems like we can avoid using `int8_t`

spot_driver/src/api/default_time_sync_api.cpp
mschweig: Yeah you're right, will change that!

spot_driver/include/spot_driver/api/default_time_sync_api.hpp
```diff
@@ -36,5 +37,6 @@ class DefaultTimeSyncApi : public TimeSyncApi {
 
  private:
   std::shared_ptr<::bosdyn::client::TimeSyncThread> time_sync_thread_;
+  int8_t timesync_timeout_;
```
jcarpinelli-bdai, line 40: If you follow @bjin-bdai's [suggestion](https://github.com/bdaiinstitute/spot_ros2/pull/552/files#r1913468357) above, this and other instances could match:

```suggestion
  std::chrono::miliseconds timesync_timeout_;
```

spot_driver/test/src/test_parameter_interface.cpp
```diff
@@ -197,6 +197,8 @@ TEST_F(RclcppParameterInterfaceEnvVarTest, GetSpotConfigFromParameters) {
   node_->declare_parameter("publish_depth_registered", publish_depth_registered_images_parameter);
   constexpr auto tf_root_parameter = "body";
   node_->declare_parameter("tf_root", tf_root_parameter);
+  constexpr auto timesync_timeout_parameter = 42;
+  node_->declare_parameter("timesync_timeout", timesync_timeout_parameter);
```
jcarpinelli-bdai, line 201: Nice test addition!

jcarpinelli-bdai (commented):
Looks good to me! I'm learning this repository myself, so holding off approval because I'm not qualified to issue it. Thank you for the contribution!

spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
```diff
@@ -196,6 +198,11 @@ bool RclcppParameterInterface::getGripperless() const {
   return declareAndGetParameter<bool>(node_, kParameterNameGripperless, kDefaultGripperless);
 }
 
+std::chrono::seconds RclcppParameterInterface::getTimeSyncTimeout() const {
+  int timeout_seconds = declareAndGetParameter<int>(node_, kParameterTimeSyncTimeout, kDefaultTimeSyncTimeout.count());
```
bjin-bdai, line 202:
```suggestion
  int timeout_seconds = declareAndGetParameter<int>(node_, kParameterTimeSyncTimeout, std::chrono::duration_cast<std::chrono::seconds>(kDefaultTimeSyncTimeout).count());
```

this gets slightly dangerous if someone decided to change `kDefaultTimeSyncTimeout` to a type like `std::chrono::milliseconds`. Without the duration cast to guarantee the type is seconds, the `count()` may return a number that's not meant for seconds (and the compiler won't catch it)

spot_driver/include/spot_driver/api/default_spot_api.hpp
```diff
@@ -19,7 +20,7 @@ namespace spot_ros2 {
 
 class DefaultSpotApi : public SpotApi {
  public:
-  explicit DefaultSpotApi(const std::string& sdk_client_name,
+  explicit DefaultSpotApi(const std::string& sdk_client_name, const std::chrono::seconds& timesync_timeout,
```
bjin-bdai, line 23:
```suggestion
  explicit DefaultSpotApi(const std::string& sdk_client_name, const std::chrono::seconds timesync_timeout,
```

nit: there's some inconsistency in the PR with the input parameter being a reference or not. I think generally for `chrono` duration types, there isn't a performance gain from taking a reference since the memory footprint of a duration object is the same size as a reference.

mschweig: oops, yeah, doesn't matter in this case. thanks, changed it.

mschweig:
> I have one comment about a potential risk with the `chrono` usage, but otherwise LGTM

Thanks, added your suggestions and tested again. Should be ready to merge.

khughes-bdai:
### Merge activity

* **Jan 14, 3:32 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/552).

PR #562, opened 2025-01-22
[N/A] Add crouching stand service

llee-bdai:
## Change Overview
Adding in a service to crouch (stand at the lowest height possible) Spot. This is needed to increase the reachable space closer to the ground :)

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -470,6 +470,12 @@ class SpotROS(Node):
         )
         self.create_service(
             Trigger,
+            "crouch",
+            lambda request, response: self.service_wrapper("crouch", self.handle_crouch, request, response),
+            callback_group=self.group,
+        )
+        self.create_service(
+            Trigger,
             "rollover",
             lambda request, response: self.service_wrapper("rollover", self.handle_rollover, request, response),
             callback_group=self.group,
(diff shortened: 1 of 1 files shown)


khughes-bdai:
I'm good with this change but just curious if you tried using the `/body_pose` topic? I haven't tried it on robot but it supposedly lets you move the height + orientation of the body at once. https://github.com/bdaiinstitute/spot_ros2/blob/24e28861481641eb007f936a7e1f475508b3a8f4/spot_driver/spot_driver/spot_ros2.py#L2506

tcappellari-bdai (approved):
im good with this addition

llee-bdai:
@khughes-bdai it does work with `/body_pose` if you publish a Pose but I kind of wanted a trigger service for this. But I also completely understand if this is just more fluff that isn't needed, in which case I will use `/body_pose`

I'm not going to die on this `crouch` hill, just makes my life easier :)

khughes-bdai (approved):
LGTM, good to know that body pose still works!

scastro-bdai (approved):
LGTM -- wondering if it makes any logical sense to change this to a `SetBool` so you can set crouch to be true/false, but maybe this instead is a bunch of trigger services for e.g., stand, sit, crouch, etc.?

It also may be too late to consider this, but is there also an option to have a single service with a string payload that can triage the string contents to do things like sit, stand, etc.?

llee-bdai:
@scastro-bdai the reason I kept it as a `Trigger` here is mostly to follow the convention of how the Spot driver has so far been written.

On the behavior stack side (where I intend to use this), it will be a string

khughes-bdai (approved):

tcappellari-bdai (approved):

PR #567, opened 2025-01-31
[SW-1891] Make Spot hardware interface polite around leases 

mhidalgo-bdai:
## Change Overview

Safe concurrent use of `spot_ros2_control` and `spot_driver` will require cooperation and coordination. This patch is a step towards both around leases. `spot_ros2_control` should not forcibly take leases, only a human should (e.g. using the Spot tablet), nor should it attempt to acquire a lease unless instructed to. 

## Testing Done

`spot_ros2_control` demos tested on Opal (thanks @khughes-bdai !)

### DIFF
--- a/spot_hardware_interface/src/spot_hardware_interface.cpp
+++ b/spot_hardware_interface/src/spot_hardware_interface.cpp
@@ -379,7 +379,7 @@ bool SpotHardware::get_lease() {
   }
   lease_client_ = lease_client_resp.response;
   // Then acquire the lease for the body.
-  const auto lease_res = lease_client_->TakeLease("body");
+  const auto lease_res = lease_client_->AcquireLease("body");
   if (!lease_res) {
     RCLCPP_ERROR(rclcpp::get_logger("SpotHardware"), "Could not acquire body lease");
     return false;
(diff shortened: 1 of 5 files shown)


spot_hardware_interface/src/spot_hardware_interface.cpp
```diff
@@ -379,7 +379,7 @@ bool SpotHardware::get_lease() {
   }
   lease_client_ = lease_client_resp.response;
   // Then acquire the lease for the body.
-  const auto lease_res = lease_client_->TakeLease("body");
+  const auto lease_res = lease_client_->AcquireLease("body");
```
khughes-bdai, line 382: nit: Can we mention somewhere in the readme that you have to release control with the tablet before launching?

khughes-bdai (approved):

spot_hardware_interface/src/spot_hardware_interface.cpp
mhidalgo-bdai: Done in https://github.com/bdaiinstitute/spot_ros2/pull/567/commits/a7412de1cd48a0055ab720e95ed15e0015416930.

PR #568, opened 2025-02-03
[SW-1890] changed take_lease_handler to acquire lease if possible, not force take

tcappellari-bdai:
## Change Overview

`spot_driver` will acquire the lease (if available) instead of forcefully taking it. If lease is not available, print an error message. Tablet can still take the lease away but needs to be manually released from the tablet for the driver to acquire it.

Related to https://github.com/bdaiinstitute/spot_wrapper/pull/154

## Testing Done

Tested on robot with a few simple trigger services

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -226,7 +226,7 @@ class SpotROS(Node):
         self.declare_parameter("auto_power_on", False)
         self.declare_parameter("auto_stand", False)
 
-        self.declare_parameter("use_take_lease", True)
+        self.declare_parameter("use_take_lease", False)
         self.declare_parameter("get_lease_on_action", True)
         self.declare_parameter("continually_try_stand", False)
 
@@ -1000,13 +1000,13 @@ class SpotROS(Node):
             response.message = "spot_ros2 is running in mock mode."
             return response
 
-        have_new_lease, lease = self.spot_wrapper.takeLease()
-        if have_new_lease:
+        try:
+            lease = self.spot_wrapper.getLease() # acquires lease if available, not forcefully take
             response.success = True
             response.message = str(lease.lease_proto)
-        else:
+        except Exception as e:
             response.success = False
-            response.message = ""
+            response.message = str(e)
 
         return response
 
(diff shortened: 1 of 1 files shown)


spot_driver/spot_driver/spot_ros2.py
```diff
@@ -226,7 +226,7 @@ def __init__(self, parameter_list: Optional[typing.List[Parameter]] = None, **kw
         self.declare_parameter("auto_power_on", False)
         self.declare_parameter("auto_stand", False)
 
-        self.declare_parameter("use_take_lease", True)
+        self.declare_parameter("use_take_lease", False)
```
mhidalgo-bdai, line 229: @tcappellari-bdai meta: if we keep this parameter (e.g. for backwards compatibility), I believe we should at least warn the user of the implications of enabling it.

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -1000,13 +1000,13 @@ def take_lease_callback(self, request: Trigger.Request, response: Trigger.Respon
             response.message = "spot_ros2 is running in mock mode."
             return response
 
-        have_new_lease, lease = self.spot_wrapper.takeLease()
-        if have_new_lease:
+        try:
+            lease = self.spot_wrapper.getLease()  # acquires lease if available, not forcefully take
```
mhidalgo-bdai, line 1004: @tcappellari-bdai meta: is this enough, error handling wise? There are APIs down in `spot_wrapper` that acquire/take leases and do not clearly propagate errors. Crashing the driver or failing elsewhere is not a graceful behavior.

tcappellari-bdai: thats a fair point - unsure best way to go about dealing with that

mhidalgo-bdai: Hmm, how about working our way upwards starting at `spot_wrapper`? Make sure bad lease errors are handled, then do the same in the `spot_driver`. We have a number of execution paths with auto-claiming logic, so we'd have to validate those will behave if lease acquisition fails.

tcappellari-bdai:
not necessary due to design change

PR #572, opened 2025-02-06
Add Dockerfile for ARM

mschweig:
## Change Overview

Created a Dockerfile.arm for ARM deployment as the current Dockerfile does not support ARM architecture. Useful for deployment and decoupled setup with several machines. Thought it could be useful for RAI. 

## Testing Done

Tested on NVIDIA Jetson ORIN (attached to robot via ethernet) with docker-compose (also included for demo purpose)


docker-compose.yaml
```diff
@@ -0,0 +1,19 @@
+services:
+  spot-ros2-driver:
+    image: spot-ros2-driver-build
```
mhidalgo-bdai, line 3: @mschweig hmm, where is this image coming from?

docker-compose.yaml
```diff
@@ -0,0 +1,19 @@
...
+    privileged: true
+    network_mode: "host"
+    environment:
+      - RMW_IMPLEMENTATION=rmw_zenoh_cpp
```
mhidalgo-bdai, line 7: @mschweig hmm, this would be news for this repository. We have not done enough testing with Zenoh. Any insights you may want to share?

docker-compose.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - ZENOH_ROUTER_CONFIG_URI=/ros_ws/src/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
+    volumes:
+      - ./timon.yaml:/ros_ws/src/robot.yaml
+      - ./DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5:/ros_ws/src/DEFAULT_RMW_ZENOH_ROUTER_CONFIG.json5
```
mhidalgo-bdai, line 11: @mschweig I sense this patch is incomplete, where are these files? Cannot bind mount without them.

Dockerfile.arm
```diff
@@ -0,0 +1,52 @@
+# Use official Ubuntu 22.04 base image
+FROM ubuntu:22.04
```
mhidalgo-bdai, line 2: @mschweig meta: if we are relying on multi-platform images, and not specifying any platform, why not have a single `Dockerfile` that relies on `$(dpkg --print-architecture)` to configure `install_spot_ros2.sh`? I don't see any fundamental difference when comparing this to https://github.com/bdaiinstitute/spot_ros2/blob/main/Dockerfile.

mhidalgo-bdai (changes requested):
First pass

docker-compose.yaml
mschweig: @mhidalgo-bdai added docker-compose just for demo. Just wanted to discuss if an ARM image for deployment is useful for this repo :) Ofc this one needs changes.

docker-compose.yaml
mschweig: @mhidalgo-bdai I am currently testing alot with zenoh, thats why I made this image so I can run long-term tests. I happy to share the results in a while.

Dockerfile.arm
mschweig: @mhidalgo-bdai Sure, we could use `$(dpkg --print-architecture)` to configure `install_spot_ros2.sh` but the main difference to the existing Dockerfile is, that the base image `osrf/ros:humble-desktop-full` is linux/amd64 (https://hub.docker.com/r/osrf/ros/tags). 

Because of that I built it on ubuntu:22.04 and installed humble manually. 

Alternative could be: https://hub.docker.com/r/arm64v8/ros/

mhidalgo-bdai: Yeap. The difference is trivial to address though. We can either install ROS 2 manually for both architectures or use https://hub.docker.com/_/ros images.

mhidalgo-bdai: > Should I include a working docker-compose?

That'd be nice. @tcappellari-bdai @khughes-bdai for feedback.

khughes-bdai: That sounds good to me :+1:

mschweig: Hi @khughes-bdai @mhidalgo-bdai, 

changed the Dockerfile and install script to support both architectures. Added a working template docker-compose. 

Added rmw_zenoh as _experimental_ argument. If you don't like that I can remove that :)

Otherwise, I would also add to `README` how to properly configure `rmw_zenoh`.

mschweig: Alright, I can add that. 

I just successfully built it on arm64 and amd64. Had a few issues with passing the `arch` parameter, but now it works.

Dockerfile
```diff
@@ -1,31 +1,53 @@
...
-    python3-rosdep \
-    libpython3-dev && \
-    rm -rf /var/lib/apt/lists/*
-
-# Create ROS workspace
+    wget software-properties-common lsb-release gnupg2 \
+    python3-colcon-common-extensions python3-pip python3-pybind11 \
+    python3-pytest-cov python3-tk python3-rosdep python3-colcon-mixin \
+    python3-rosinstall-generator python3-vcstool libpython3-dev \
+    ros-humble-ros-base ros-dev-tools \
```
mhidalgo-bdai, line 33: @mschweig consider keeping the one dependency per line format -- it makes it easier for the human to parse the list.

docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+services:
+  spot-ros2-driver:
+    image: spot_ros2:latest
+    volumes:
+      - ./spot_driver/config/spot_ros_example.yaml:/ros_ws/src/robot.yaml
+    command: >
+      bash -c "
+        source /opt/ros/humble/setup.bash && 
+        source /ros_ws/install/setup.bash && 
+        ros2 launch spot_driver spot_driver.launch.py config_file:=/ros_ws/src/robot.yaml spot_name:=timon launch_rviz:=False
```
mhidalgo-bdai, line 10: @mschweig instead of hardcoding `spot_name`, we could pick it up from the environment.

Dockerfile
```diff
@@ -1,31 +1,53 @@
...
-# Run install script
-RUN /ros_ws/src/install_spot_ros2.sh
+# Run install script and pass in the architecture
+RUN ARCH=$(dpkg --print-architecture) && echo "Building driver with $ARCH" && /ros_ws/src/install_spot_ros2.sh --$ARCH
 
 # Build packages with Colcon
 WORKDIR /ros_ws/
-RUN . /opt/ros/humble/setup.sh && \
-    colcon build --symlink-install
+RUN /bin/bash -c "source /opt/ros/humble/setup.sh && colcon build --symlink-install"
```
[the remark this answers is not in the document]
mschweig: @mhidalgo-bdai I tried to follow the official ROS 2 Humble installation documentation. They use `source` but yeah, in this case, `source` and `.` do the same thing. I can use the old command if you prefer it. :)

`/bin/bash -c ` can be omitted anyway as I set it before as default. Will change that! Thanks for the hint.

Dockerfile
```diff
@@ -1,31 +1,53 @@
...
+RUN rosdep init && rosdep update
+
 # Clone driver code
 RUN git clone https://github.com/bdaiinstitute/spot_ros2.git .
 RUN git submodule update --init --recursive
 
-# Run install script
-RUN /ros_ws/src/install_spot_ros2.sh
+# Run install script and pass in the architecture
+RUN ARCH=$(dpkg --print-architecture) && echo "Building driver with $ARCH" && /ros_ws/src/install_spot_ros2.sh --$ARCH
```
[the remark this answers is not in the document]
mschweig: @mhidalgo-bdai  Right, I am going to add a verification to `install_spot_ros2.sh` I think it fits better there.

docker-compose.yaml
mschweig: @mhidalgo-bdai  hmm, but I'd suggest to pick `spot_name` from env as a default in the launch file? or introduce a new `ros_parameter` in `spot_ros_example.yaml` ?

mhidalgo-bdai: @mschweig hmm, I'd say we keep the launch files the way they are. In this case, an environment variable is just a mechanism to configure the `docker-compose` service. If you do `docker-compose run` you could also use command line arguments. 

I don't mind strongly how, I just wouldn't hardcode some robot name in the `docker-compose.yaml` file.

mschweig: @mhidalgo-bdai, @khughes-bdai  all done now :)

mhidalgo-bdai (approved):
LGTM though I don't have an `arm64` box at hand.

khughes-bdai (approved):

PR #578, opened 2025-02-12
[SW-957] make decorator for trigger services

tcappellari-bdai:
## Change Overview

Made a descriptor class for trigger services to reduce boilerplate + copy/paste code in trigger service handlers

## Testing Done

Tested all trigger services on robot. Found a bug with estop handling within `spot_wrapper` (will make a new ticket to track this but out of scope for this PR)

### DIFF
--- a/spot_driver/spot_driver/ros_helpers.py
+++ b/spot_driver/spot_driver/ros_helpers.py
@@ -1,6 +1,7 @@
 import os
 import time
 from typing import Any, Callable, List, Optional, Tuple, Union
+from dataclasses import dataclass
 
 import builtin_interfaces.msg
 import cv2
@@ -20,6 +21,7 @@ from google.protobuf.timestamp_pb2 import Timestamp
 from rclpy.node import Node
 from sensor_msgs.msg import CameraInfo, CompressedImage, Image
 from tf2_msgs.msg import TFMessage
+from std_srvs.srv import SetBool, Trigger
 
 from spot_wrapper.wrapper import SpotWrapper
 
@@ -328,3 +330,16 @@ def lookup_a_tform_b(
                 raise e
             time.sleep(0.01)
     return None
+
+@dataclass
+class TriggerServiceDescriptor:
+    service_name: str
+    def __get__(self, obj, type=None):
+        def handler(request: Trigger.Request, response: Trigger.Response):
+            if obj.spot_wrapper is None:
+                response.success = False
+                response.message = "Spot wrapper is undefined"
+                return response
+            response.success, response.message = getattr(obj.spot_wrapper, self.service_name)()
+            return response
+        return handler
--- a/spot_driver/spot_driver/ros_helpers.py
+++ b/spot_driver/spot_driver/ros_helpers.py
@@ -1,6 +1,5 @@
 import os
 import time
-from dataclasses import dataclass
 from typing import Any, Callable, List, Optional, Tuple, Union
 
 import builtin_interfaces.msg
@@ -18,6 +17,7 @@ from builtin_interfaces.msg import Time
 from cv_bridge import CvBridge
 from geometry_msgs.msg import TransformStamped
 from google.protobuf.timestamp_pb2 import Timestamp
+from rclpy.callback_groups import CallbackGroup
 from rclpy.node import Node
 from sensor_msgs.msg import CameraInfo, CompressedImage, Image
 from std_srvs.srv import Trigger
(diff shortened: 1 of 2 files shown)


spot_driver/spot_driver/ros_helpers.py
```diff
@@ -328,3 +330,41 @@ def lookup_a_tform_b(
...
     return None
+
+
+@dataclass
+class TriggerServiceDescriptor:
+    """A descriptor for calling callbacks for trigger services"""
+
+    service_path: str
+    args: tuple = ()
+    kwargs: dict = None
```
amessing-bdai, line 341:
```suggestion
    kwargs: Optional[dict] = None
```

spot_driver/spot_driver/spot_ros2.py
```diff
@@ -1110,186 +1102,34 @@ def service_wrapper(
...
-
-    def handle_sit(self, request: Trigger.Request, response: Trigger.Response) -> Trigger.Response:
-        """ROS service handler for the sit service"""
-        if self.spot_wrapper is None:
-            response.success = False
-            response.message = "Spot wrapper is undefined"
-            return response
-        response.success, response.message = self.spot_wrapper.sit()
-        return response
+    handle_claim = TriggerServiceDescriptor("claim")
```
khughes-bdai, line 1105: would it be easier to pass the function it's calling rather than the string? (like `self.spot_wrapper.claim` instead of `"claim"`, then you wouldn't have to parse the string )

jcarpinelli-bdai (approved):

spot_driver/spot_driver/ros_helpers.py
```diff
@@ -328,3 +330,46 @@ def lookup_a_tform_b(
...
+class TriggerServiceWrapper:  # TODO: caching of spot_ros might be funky with multiple instances up
+    """A descriptor for calling callbacks for trigger services"""
+
+    service_func: Callable[[SpotWrapper], Trigger.Response]
+    service_name: str
+    args: Optional[tuple] = ()
+    kwargs: Optional[dict | None] = None
+
+    def __init__(
+        self, service_func: Callable, service_name: str, *args: Optional[Any], **kwargs: Optional[Any | None]
```
bjin-bdai, line 344:
```suggestion
        self, service_func: Callable, service_name: str, *args: Optional[Any], **kwargs: Optional[Any]
```

nit: does `Any` cover the `None` type?

spot_driver/spot_driver/ros_helpers.py
```diff
@@ -328,3 +330,46 @@ def lookup_a_tform_b(
...
+    ) -> None:
+        self.service_func = service_func
+        self.service_name = service_name
+        self.args = args
+        self.kwargs = kwargs or {}
+
+    def create_service(self, obj: Any, group: CallbackGroup) -> None:
```
bjin-bdai, line 351: `obj: Any` is pretty vague. Based on the naming as well it seems like this `obj` type is limited to a set of Spot objects?

bjin-bdai (commented):
Variable naming for `obj: Any` might be a little too vague

spot_driver/spot_driver/ros_helpers.py
tcappellari-bdai: ya its supposed to be SpotRos object but that would cause a circular dependency. Was unsure best way around that and keep mypy happy

khughes-bdai (approved):

spot_driver/spot_driver/ros_helpers.py
```diff
@@ -328,3 +330,46 @@ def lookup_a_tform_b(
...
+        self.spot_ros = obj
+        obj.create_service(Trigger, self.service_name, self.callback, callback_group=group)
+
+    def callback(self, request: Trigger.Request, response: Trigger.Response) -> Trigger.Response:
+        if self.spot_ros.mock:
```
bjin-bdai, line 356: If this callback explcitly depends on SpotRos object, then does it make sense that `TriggerServiceWrapper` is its own class at all? Could the code duplication also be solved by moving these function to member functions of `SpotRos`? (and the member variables to this class can be stored as a dictionary, or wrapped in a function wrapper tool?)

The circular dependency comment suggests that there is a different way to implement this... and I'm thinking if this class specifically only works on `SpotRos` object, then they are functions/properties that belong to `SpotRos` class

tcappellari-bdai: ya it makes sense to put it in spotROS. I just wanted to reduce the amount of code in it since its already m a s s i v e

khughes-bdai (approved):

tcappellari-bdai:
### Merge activity

* **Feb 14, 3:10 PM EST**: [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/578) rebased this pull request as part of a merge.

PR #579, opened 2025-02-12
Fix false trajectory success when blocked by obstacle

gaddison-bdai:
## Change Overview

Just changed some logic inside of `handle_trajectory()` member function to properly handle the new flags/logic created in the corresponding `spot_wrapper` branch. Going to keep this as a draft until the corresponding branch in `spot_wrapper` is merged to main.

Depends on https://github.com/bdaiinstitute/spot_wrapper/pull/156

## Testing Done

Tested on Rex in spot lab. Confirmed that it will report "Trajectory complete: failed to reach goal" if it is unable to reach the goal. Previously, it would instead say "Reached goal" and success. Trajectory duration time out fail also tested and still sends proper feedback and result.

### DIFF
--- a/spot_driver/spot_driver/spot_ros2.py
+++ b/spot_driver/spot_driver/spot_ros2.py
@@ -2449,15 +2449,15 @@ class SpotROS(Node):
         # rate = rclpy.Rate(10)
 
         try:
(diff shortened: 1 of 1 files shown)


tcappellari-bdai (approved):
approved once corresponding spot_wrapper PR is merged and CI passes

PR #581, opened 2025-02-14
[SW-2000] increase robot state publisher rates for faster tf lookups

khughes-bdai:
## Change Overview

The robot state publisher defaults to a TF update rate of 20 Hz, this needs to be increased to prevent slow TF lookups :) 
* spot driver publishes joint states at 50 Hz, so its TF update rate should be at least that
* spot ros2 control publishes joint states at 333 Hz, so its TF update rate should be at least that 

## Testing Done

- [x] Test spot driver and ensure publishing rates of other topics is similar to before
- [x] Test spot ros2 control and ensure publishing rates of other topics is similar to before

### DIFF
--- a/spot_ros2_control/launch/spot_ros2_control.launch.py
+++ b/spot_ros2_control/launch/spot_ros2_control.launch.py
@@ -164,12 +164,13 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
             remappings=[(f"/{tf_prefix}joint_states", f"/{tf_prefix}low_level/joint_states")],
         )
     )
+    # Joint states get published at 333 Hz, so set the TF update rate to that.
     ld.add_action(
         Node(
             package="robot_state_publisher",
             executable="robot_state_publisher",
             output="both",
-            parameters=[robot_description],
+            parameters=[robot_description, {"publish_frequency": 333.0}],
             namespace=spot_name,
             remappings=[(f"/{tf_prefix}joint_states", f"/{tf_prefix}low_level/joint_states")],
         )
--- a/spot_ros2_control/launch/spot_ros2_control.launch.py
+++ b/spot_ros2_control/launch/spot_ros2_control.launch.py
@@ -164,13 +164,14 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
             remappings=[(f"/{tf_prefix}joint_states", f"/{tf_prefix}low_level/joint_states")],
         )
     )
-    # Joint states get published at 333 Hz, so set the TF update rate to that.
+    # Publish frequency of the robot state publisher defaults to 20 Hz, resulting in slow TF lookups.
+    # By ignoring the timestamp, we publish a TF update in this node every time there is a joint state update (333 Hz).
     ld.add_action(
         Node(
             package="robot_state_publisher",
             executable="robot_state_publisher",
             output="both",
-            parameters=[robot_description, {"publish_frequency": 333.0}],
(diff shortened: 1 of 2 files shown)


khughes-bdai:
* **#581** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/581?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/581?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -164,12 +164,13 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
...
         )
     )
+    # Joint states get published at 333 Hz, so set the TF update rate to that.
     ld.add_action(
         Node(
             package="robot_state_publisher",
             executable="robot_state_publisher",
             output="both",
-            parameters=[robot_description],
+            parameters=[robot_description, {"publish_frequency": 333.0}],
```
mhidalgo-bdai, line 173: @khughes-bdai meta: this is true so long as the controllers' update rate is set to 333 Hz. I wonder if it wouldn't be better to define this in one place and consume it from elsewhere (e.g. tapping into parameter files on launch or using substitutions in parameter files to interpolate launch configurations).

Same above.

khughes-bdai: Yeah, I agree that would be cleaner, i think it would require some extra re-factoring. the 50 hz joint state rate for the driver is hardcoded [here](https://github.com/bdaiinstitute/spot_ros2/blob/c8dc94f21f8b2578e874fed917bd2bd14146d044/spot_driver/src/robot_state/state_publisher.cpp#L13) and it would need to get turned into a parameter, and the ros2 control update rate is set in the config file.

although if i understand the implementation of robot state publisher correctly, the tf gets updated at most every time there is a joint state update (as [it is broadcasting tf in a joint state callback](https://github.com/ros/robot_state_publisher/blob/be491624cc808fe2bb8689c630748c09608400eb/src/robot_state_publisher.cpp#L333)). so as long as `publish_frequency` >= joint state update rate, these two topics should stay in sync?

mhidalgo-bdai: > the tf gets updated at most every time there is a joint state update so as long as publish_frequency >= joint state update rate

That's right. In fact, there is an `ignore_timestamp` parameter for the publisher to broadcast on every joint state update. Never used it before, may come in handy here.

khughes-bdai: Oh i see, yeah that would simplify things. I'll test that param on robot to make sure that everything still works as expected

khughes-bdai: I changed these launchfiles to use ignore_timestamp, and it seems to do what I want on robot! Thanks for the suggestion

khughes-bdai:
### Merge activity

* **Feb 28, 3:10 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/581).

PR #598, opened 2025-02-27
[SW-2051] expose foot contact states to state streaming handler

tcappellari-bdai:
## Change Overview

expose foot contact states in state streaming handler

## Testing Done

### DIFF
--- a/spot_hardware_interface/include/spot_hardware_interface/spot_hardware_interface.hpp
+++ b/spot_hardware_interface/include/spot_hardware_interface/spot_hardware_interface.hpp
@@ -210,7 +210,6 @@ class SpotHardware : public hardware_interface::SystemInterface {
   // Holds IMU data for the robot received from the BD SDK
   ImuStates imu_states_;
   // Holds foot states received from the BD SDK
-  // ::bosdyn::api::FootState::Contact foot_states_;
   std::vector<int> foot_states_;
 
   // Thread for reading the state of the robot.
--- a/spot_hardware_interface/include/spot_hardware_interface/spot_hardware_interface.hpp
+++ b/spot_hardware_interface/include/spot_hardware_interface/spot_hardware_interface.hpp
@@ -107,11 +107,7 @@ class StateStreamingHandler {
       CONTACT_MADE	1	The foot is currently in contact with the ground.
       CONTACT_LOST	2	The foot is not in contact with the ground.
    */
-  void get_states(JointStates& joint_states, ImuStates& imu_states);
-  /**
-   * @brief Reset internal state.
-   */
-  void reset();
+  void get_foot_states(std::vector<int>& foot_states);
 
  private:
   // Stores the current position, velocity, and load of the robot's joints.
@@ -126,6 +122,7 @@ class StateStreamingHandler {
   std::vector<double> imu_odom_rot_quaternion_;
   // store the current foot contact states
   std::vector<int> current_foot_state_;
+  static constexpr size_t nfeet_ = 4;
   // responsible for ensuring read/writes of joint states do not happen at the same time.
   std::mutex mutex_;
 };
@@ -175,7 +172,7 @@ class SpotHardware : public hardware_interface::SystemInterface {
   // The 3 state interfaces are position, velocity, and effort.
   static constexpr size_t state_interfaces_per_joint_ = 3;
   size_t njoints_;
-  size_t nfeet = 4;
+  static constexpr size_t nfeet_ = 4;
 
   // Login info
   std::string hostname_;
--- a/spot_hardware_interface/include/spot_hardware_interface/spot_hardware_interface.hpp
+++ b/spot_hardware_interface/include/spot_hardware_interface/spot_hardware_interface.hpp
@@ -98,7 +98,7 @@ class StateStreamingHandler {
    */
   void handle_state_streaming(::bosdyn::api::RobotStateStreamResponse& robot_state);
   /**
-   * @brief Get structs of the current joint states and IMU data from the robot.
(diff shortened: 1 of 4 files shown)


spot_hardware_interface/include/spot_hardware_interface/spot_hardware_interface.hpp
```diff
@@ -140,6 +149,7 @@ class SpotHardware : public hardware_interface::SystemInterface {
   // The 3 state interfaces are position, velocity, and effort.
   static constexpr size_t state_interfaces_per_joint_ = 3;
   size_t njoints_;
+  size_t nfeet = 4;
```
bjin-bdai, line 152:
```suggestion
  static constexpr size_t kNumFeet_ = 4;
```
seems like this is a constant?

spot_hardware_interface/src/spot_hardware_interface.cpp
```diff
@@ -40,6 +40,10 @@ void StateStreamingHandler::handle_state_streaming(::bosdyn::api::RobotStateStre
   current_position_.assign(position_msg.begin(), position_msg.end());
   current_velocity_.assign(velocity_msg.begin(), velocity_msg.end());
   current_load_.assign(load_msg.begin(), load_msg.end());
+  // Save current foot contact states
+  for (size_t i = 0; i < 4; i++) {
```
bjin-bdai, line 44:
```suggestion
  for (size_t i = 0; i < nfeet_; i++) {
```

spot_hardware_interface/src/spot_hardware_interface.cpp
```diff
@@ -40,6 +40,10 @@ void StateStreamingHandler::handle_state_streaming(::bosdyn::api::RobotStateStre
...
+    current_foot_state_.push_back(robot_state.contact_states(i));
```
bjin-bdai, line 45: Q: I'm slightly confused by what this vector represents. It's seem seems like a 4-element vector, but then each element is increasing from 0-3. What does 0 mean vs 3? Or is this supposed to be a vector of booleans representing "is in contact" vs "not in contact"?

jcarpinelli-bdai (approved):

khughes-bdai:
I think to get these exported as a state interface, you'll first have to modify the ros2 control tags in the urdf to add in sensor related interfaces. it's definitely a non standard sensor type so idk the best way to do it, but maybe something like this?
```
        <sensor name="feet_contacts">
            <state_interface name="foot_state.front.left"/>
            <state_interface name="foot_state.front.right"/>
            <state_interface name="foot_state.rear.left"/>
            <state_interface name="foot_state.rear.right"/>
        </sensor>
```

spot_hardware_interface/src/spot_hardware_interface.cpp
tcappellari-bdai: according to BD's documentation 0 is unknown, 1 is contact, and 2 and no contact

spot_hardware_interface/src/spot_hardware_interface.cpp
```diff
@@ -51,6 +55,12 @@ void StateStreamingHandler::get_joint_states(JointStates& joint_states) {
   joint_states.load.assign(current_load_.begin(), current_load_.end());
 }
 
+void StateStreamingHandler::get_foot_states(std::vector<int>& foot_states) {
+  // lock so that read/write doesn't happen at the same time
+  const std::lock_guard<std::mutex> lock(mutex_);
+  foot_states.assign(current_foot_state_.begin(), current_foot_state_.end());
+}
+
```
khughes-bdai, line 63: <p class='graphite__hidden'><i>[Re: lines +49 to +63]</i></p>
I'm wondering if this could just be consolidated with the above `get_joint_states` function so we could have just one generic `get_states` function? then we can be sure that the joint states and the foot states (and whatever other states we want to expose in the future) are all obtained at the same time
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/598?utm_source=unchanged-line-comment">Graphite</a>.</p>

tcappellari-bdai: I had split them up for organizational purposes but it makes sense to consolidate into a single method too. I think I prefer that too

spot_hardware_interface/src/spot_hardware_interface.cpp
bjin-bdai: misread the code, lgtm

bjin-bdai (approved):

spot_hardware_interface/include/spot_hardware_interface/spot_hardware_interface.hpp
```diff
@@ -98,16 +98,16 @@ class StateStreamingHandler {
...
+   *  CONTACT_UNKNOWN	0	Unknown contact. Do not use.
+      CONTACT_MADE	1	The foot is currently in contact with the ground.
+      CONTACT_LOST	2	The foot is not in contact with the ground.
    */
-  void get_states(JointStates& joint_states, ImuStates& imu_states);
-  /**
-   * @brief Reset internal state.
-   */
-  void reset();
+  void get_states(JointStates& joint_states, ImuStates& imu_states, std::vector<int>& foot_states);
```
robodreamer, line 110: A quick question, is there any code that used the `get_joint_states` API that has to be also updated with this change? In other words, have we checked if this would be a breaking change?

tcappellari-bdai: The `get_joint_states` was only used internally for the hardware interface (as far as I'm aware of) to load information received from the state streamer so it shouldn't be a breaking change

PR #618, opened 2025-03-07
[SW-2229] clean up autogeneration of ros2 control config files

khughes-bdai:
## Change Overview

## Testing Done

### DIFF
--- a/spot_ros2_control/launch/spot_ros2_control.launch.py
+++ b/spot_ros2_control/launch/spot_ros2_control.launch.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2024 Boston Dynamics AI Institute LLC. All rights reserved.
+# Copyright (c) 2024-2025 Boston Dynamics AI Institute LLC. All rights reserved.
 import os
 from tempfile import NamedTemporaryFile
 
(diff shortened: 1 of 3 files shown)


khughes-bdai:
* **#618** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/618?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

mhidalgo-bdai (commented):
@khughes-bdai this is cleaner code, but I can't help thinking it makes (re)configuration harder. Do we really need this much prefixing? We can do blanket node namespacing from `launch` and the only other thing that needs prefixing are the joint names for `robot_state_publisher` to broadcast `tf` properly (and we can solve that using a [`frame_prefix`](https://github.com/ros/robot_state_publisher/blob/8b5af4f31e0f754c7c262eb487ccd3956cb788cc/src/robot_state_publisher.cpp#L257)).

jcarpinelli-bdai (approved):

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -160,11 +193,17 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
     )
     robot_description = {"robot_description": robot_urdf}
 
+    try:
+        # convert update rate to an int, and clamp between min and max.
+        update_rate = max(MIN_UPDATE_RATE_HZ, min(int(update_rate), MAX_UPDATE_RATE_HZ))
```
bjin-bdai, line 198: nit, would using something like `np.clip` be better here?

spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -160,11 +193,17 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
...
+    except ValueError:
+        print(f"Update rate '{update_rate}' could not be converted to an int! Defaulting to '{MAX_UPDATE_RATE_HZ}'.")
+        update_rate = MAX_UPDATE_RATE_HZ
```
bjin-bdai, line 201: rather than automatically correcting a user error, I feel it's better to return some kind of error with a message that indicates the acceptable bounds. This way users who put the wrong update rate are never caught by surprise

mhidalgo-bdai:
> Don't the joint names also need to be namespaced for the joint state broadcaster, otherwise the joint state message will not match the joints/links given in the main URDF?

Yes, if URDF joint and link names are prefixed. Do we need them prefixed? We are namespacing the `/robot_description` topic too after all. 

> I think the main reason I prefixed the hardware components was to try to match the other prefixing we do in the high level driver. 

I know. And it makes sense in that context. It just feels like it's getting out of hand. It might well out of scope for this patch (and I won't block on this), but perhaps multi-robot support could use some thought. Just my 2 cents.

spot_ros2_control/launch/spot_ros2_control.launch.py
khughes-bdai: i got rid of the update rate configuration in this PR, wasn't necessary to include here

tcappellari-bdai (approved):

khughes-bdai:
> > Don't the joint names also need to be namespaced for the joint state broadcaster, otherwise the joint state message will not match the joints/links given in the main URDF?
> 
> Yes, if URDF joint and link names are prefixed. Do we need them prefixed? We are namespacing the `/robot_description` topic too after all. 
> 
> > I think the main reason I prefixed the hardware components was to try to match the other prefixing we do in the high level driver. 
> 
> I know. And it makes sense in that context. It just feels like it's getting out of hand. It might well out of scope for this patch (and I won't block on this), but perhaps multi-robot support could use some thought. Just my 2 cents.

@mhidalgo-bdai got sidetracked with other stuff over the last few days, just coming back to this PR. I agree that it is overkill to be doing this much prefixing but I don't want to change that here since that would be a breaking change with a lot of fallout.  For the time being, it seems easier to maintain generating the config like this. If we do proper releases/versioning of this repo, how we deal with prefixing would definitely be something i'd like to change.

khughes-bdai:
### Merge activity

* **Mar 17, 3:06 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/618).

PR #619, opened 2025-03-10
Use `rosdep` to install ROS 2 dependencies

Imaniac230:
## Change Overview

Just a small thing. I refactored the ROS dependency installation steps in `install_spot_ros2.sh` to use the `rosdep` utility instead of having to add the manual install commands. I made an assumption that `rosdep` might not be installed on the machine at all, so it will also attempt to install and initialize.

There were 3 packages that are not included in any `package.xml` definitions:
* `tf_transformations`
* `ros2_control`
* `ros2_controllers`

All spot packages are building locally on my machine and the driver launches without having these three installed, so I didn't proactively include them anywhere. However, we don't have the joint-level API license, so I don't think I'll be able to test if any of the `spot_ros2_control` nodes don't need some of those libraries at runtime.

Unrelated: I also added the new `spot_common` package to build and test steps for CI.

## Testing Done

1. Removed all ROS dependencies that were listed explicitly in old `install_spot_ros2.sh`.
2. Ran the new version and observed that all missing ROS 2 packages are being installed through `rosdep`.
3. `colcon build` all packages and verify that everything builds.
4. Launch the spot driver and verify that all works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,13 +106,13 @@ jobs:
       - name: Build packages
         run: |
           source /opt/ros/$ROS_DISTRO/setup.bash
-          colcon build --symlink-install --packages-up-to spot_driver spot_description spot_msgs spot_examples spot_ros2_control --cmake-args -DCMAKE_CXX_FLAGS="--coverage"
+          colcon build --symlink-install --packages-up-to spot_driver spot_description spot_msgs spot_examples spot_ros2_control spot_common --cmake-args -DCMAKE_CXX_FLAGS="--coverage"
         working-directory: ${{ github.workspace }}/../../
 
       - name: Test non-spot-driver packages
         run: |
           source install/setup.bash
-          colcon test --event-handlers console_direct+ --packages-select spot_description spot_examples spot_msgs
+          colcon test --event-handlers console_direct+ --packages-select spot_description spot_examples spot_msgs spot_common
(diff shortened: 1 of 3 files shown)


khughes-bdai:
`ros2_control` and `ros2_controllers` dependencies are needed by the spot ros2 control stack -- they should probably get added to the `spot_ros2_control` package.xml. It doesn't seem like the tf_transformations package is used anymore

Imaniac230:
Added them to `spot_ros2_control`.

I was thinking if it wouldn't make more sense to place `ros2_controllers` into `spot_controllers`, just judging by the names. But I don't have any strong opinions on that and the installation will work either way.

khughes-bdai:
yeah technically `ros2_controllers` is needed by both `spot_controllers` and `spot_ros2_control`, and `ros2_control` is needed by both `spot_hardware_interface` and `spot_ros2_control`. I think it's fine to leave this as-is here, as `spot_ros2_control` is designed to be the main entrypoint into the stack

khughes-bdai (approved):
Tested with our dockerfile (which calls the install script) and everything looks like it is pulled in and building correctly. Thanks!

khughes-bdai:
### Merge activity

* **Mar 10, 2:59 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_ros2/619).

Issue #621, opened 2025-03-12
Build fails on latest main


mschweig:
Hi @khughes-bdai 

build is failing due to dependency issues. Probably introduced with https://github.com/bdaiinstitute/spot_ros2/pull/619, did you already see this?

You need build with `--no-cache` or clean docker builder cache. 

> --- stderr: spot_controllers
127.9 CMake Error at CMakeLists.txt:25 (find_package):
127.9   By not providing "Findcontroller_interface.cmake" in CMAKE_MODULE_PATH this
127.9   project has asked CMake to find a package configuration file provided by
127.9   "controller_interface", but CMake did not find one.
127.9 
127.9   Could not find a package configuration file provided by
127.9   "controller_interface" with any of the following names:
127.9 
127.9     controller_interfaceConfig.cmake
127.9     controller_interface-config.cmake
127.9 
127.9   Add the installation prefix of "controller_interface" to CMAKE_PREFIX_PATH
127.9   or set "controller_interface_DIR" to a directory containing one of the
127.9   above files.  If "controller_interface" provides a separate development
127.9   package or SDK, be sure it has been installed.
127.9 
127.9 
127.9 ---
127.9 Failed   <<< spot_controllers [1.06s, exited with code 1]
127.9 Aborted  <<< spot_hardware_interface [1.05s]

Otherwise we can work on a fix!

khughes-bdai:
Hmm when trying to reproduce this I get the same errors but with `bondcpp` when trying to build `spot_hardware_interface`. Definitely seems related to the rosdep change, if you are able to work on a fix that would be greatly appreciated 😄 otherwise we can look into it here later this week

mschweig:
Alright, on it :)

mschweig:
@khughes-bdai is there a specific reason why you are caching in the ci pipeline?
        
>       - name: Build and push Docker image (may be cached)
>         uses: docker/build-push-action@v5  # https://github.com/docker/build-push-action
>         with:
>           context: .
>           file: .devcontainer/Dockerfile
>           push: true
>           tags: ${{ steps.meta.outputs.tags }}
>           labels: ${{ steps.meta.outputs.labels }}
>           cache-from: type=gha
>           cache-to: type=gha,mode=max

Imaniac230:
> This seems to be the error
> 
> ```
> updated cache in /root/.ros/rosdep/sources.cache
> WARNING: ROS_PYTHON_VERSION is unset. Defaulting to 3
[...]

PR #622, opened 2025-03-12
Fixes rosdep distro 

mschweig:
Fixes https://github.com/bdaiinstitute/spot_ros2/issues/621

Tested with Docker and `--no-cache`

Maybe it would be good to disable cache in the CI pipelines to catch these issues?


khughes-bdai (approved):
Thanks for the change! The CI pipeline actually doesn't test this specific Dockerfile/install script, it uses a more minimal one [here](https://github.com/bdaiinstitute/spot_ros2/blob/main/.devcontainer/Dockerfile). We can make a note to try to unify these in future work.

PR #629, opened 2025-03-18
Add local grid publisher

ryan-roche:
## Change Overview

This PR adds a module `/spot_local_grid` with a ROS2 node providing a publisher for Spot's local grid.

image
(Grid messages visualized in Foxglove Studio)

It is currently hard-coded to publish the `obstacle_distance` grid, but I can refactor it to take an argument for which grid to publish.

Since this was something I developed internally as part of another project, I separated it to its own module. Should you choose to merge this into the driver repo, it might make more sense to integrate it into the driver itself and have it optionally activated when using the driver launchfile, similar to the `--publish-point-clouds` argument.

## Testing Done

Being that this module is so dependent on data from the robot, I wasn't able to make any unit tests for it.

### DIFF
--- a/spot_driver/spot_driver/spot_local_grid_publisher.py
+++ b/spot_driver/spot_driver/spot_local_grid_publisher.py
@@ -1,0 +1,231 @@
+import os
+import cv2
+import rclpy
+import numpy as np
+import ros2_numpy as rnp
+from bosdyn.api import local_grid_pb2
+from bosdyn.client import create_standard_sdk
+from bosdyn.client.local_grid import LocalGridClient
+from bosdyn.client.math_helpers import SE3Pose
+from geometry_msgs.msg import Pose, Point, Quaternion
+from nav_msgs.msg import OccupancyGrid
+from rclpy.node import Node
+from sensor_msgs.msg import Image, CompressedImage
+from spot_driver.manual_conversions import se3_pose_to_ros_pose
+
+DEFAULT_LOCAL_GRID_NAME = 'obstacle_distance'
+
+
+class LocalGridPublisher(Node):
+
+    def __init__(self):
+        super().__init__('local_grid_publisher')
+        self.get_logger().info("Initializing LocalGridPublisher Node...")
+
+        # Check for robot credentials in environemnt
+        self.SPOT_IP = os.environ.get('SPOT_IP')
+        self.BOSDYN_CLIENT_USERNAME = os.environ.get('BOSDYN_CLIENT_USERNAME')
+        self.BOSDYN_CLIENT_PASSWORD = os.environ.get('BOSDYN_CLIENT_PASSWORD')
+
+        if not self.SPOT_IP or not self.BOSDYN_CLIENT_USERNAME or not self.BOSDYN_CLIENT_PASSWORD:
+            self.get_logger().error("Robot credentials not found. Ensure that the following environment variables are set: SPOT_IP, BOSDYN_CLIENT_USERNAME, BOSDYN_CLIENT_PASSWORD")
+            raise ValueError("Robot credentials not found")
+        
+        # Verify the credentials are correct
+        self.sdk = create_standard_sdk('local_grid_publisher')
+        self.robot = self.sdk.create_robot(self.SPOT_IP)
+        self.robot.authenticate(self.BOSDYN_CLIENT_USERNAME, self.BOSDYN_CLIENT_PASSWORD)   # an exception will be raised if authentication fails
+        self.get_logger().info("🔐 Robot authenticated successfully")
+        self.get_logger().info("Waiting for time sync...")
+        self.robot.time_sync.wait_for_sync()
+        self.get_logger().info("🕰 Time sync successful!")
+
+        # Create LocalGridClient
+        self.get_logger().info("Creating LocalGridClient...")
+        
+        self.local_grid_client = self.robot.ensure_client(LocalGridClient.default_service_name)
+
+        self.get_logger().info("💠 LocalGridClient created successfully!")
+
+        # Create ROS2 publisher
+        self.get_logger().info("Creating OccupancyGrid publisher...")
+        self.occupancy_grid_pub = self.create_publisher(OccupancyGrid, '/autogrammetry/local_grid', 10)
+        self.get_logger().info("📢 OccupancyGrid publisher created successfully!")
+
+        # Indicate successful initialization
+        self.get_logger().info('[✓] Spot Local Grid Publisher Node initialized')
+
+        self.first_draw_done = False
+        self.im = None
+        self.fig = None
+        self.ax = None
+
+        self.fetch_next_grid_data()
+
+    
+    def fetch_next_grid_data(self):
+        future = self.local_grid_client.get_local_grids_async([LOCAL_GRID_NAME])
+        future.add_done_callback(self.publish_grid)
+
+
+    def publish_grid(self, future):
+        """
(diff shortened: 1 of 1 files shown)


tcappellari-bdai:
Can you run `pre-commit install` then `pre-commit run --all` at the repo's root, then commit those changes to ensure the linter passes?

tcappellari-bdai:
Actually, after thinking about this a bit more, I think it would be better to have this live in the existing `spot_driver` package rather than making a new one. I also like your idea of adding an optional argument to activate it from the driver launchfile. 

Could you make those changes and also modify your node to follow our conventions for homogeneity (i.e. going through `spot_wrapper` to remove some boilerplate)? Thanks!

ryan-roche:
Where exactly in the `spot_driver` package should I put the source file for the grid publisher? Most of the nodes appear to be written in C++, but the python source directory `/spot_driver/spot_driver` doesn't have  so should I just put it at `/spot_driver/spot_driver/local_grid_pub.py`?

If I recall correctly, you _can_ have executable Nodes written in both languages within the same package, so it shouldn't be an issue that the code is in Python.

ryan-roche:
Re-synced my fork with the original- continuing in a new PR

ryan-roche:
Bumping to re-open the PR. I've tried looking into a way to add the python node to the base `spot_driver` package, but while you *can* have python files installed as part of an `ament_cmake` package, I'm trying to figure out how you add the python package dependencies.

I could try rewriting the node in C++, but I'm unfamiliar with the Spot C++ SDK so it'd take some time. The Python local grid publisher node is working as-is, so I would rather figure out how to integrate that into the driver and have rewriting it in C++ as a possible future improvement down the line. I'm sure there _is_ a way to do this- I just need to look through more of this documentation.

khughes-bdai:
@ryan-roche no need to convert to C++ -- we already have some python code in `spot_driver/spot_driver`, living side by side with the C++ code. I think the two main options for where to put this code are a) leave it as is in its own file, as you are doing now, or b) try to add the relevant code into `spot_driver/spot_driver/spot_ros2.py`, which is where almost all of our python based driver code lives. 
Ideally the second option would reduce some boilerplate (as we already create the robot object, authenticate, etc. in there, using `spot_wrapper` as a common entry point) but I'm not opposed to the first option as this node seems pretty self contained. @tcappellari-bdai thoughts?

ryan-roche:
I'll keep it in its own node- think that'll make it easier to _not_ run it unless the user explicitly asks for it with the launch arguments. Just gotta figure out how you specify package dependencies for your python nodes in an `ament_cmake` package since there isn't a `setup.py` to list them in.

ryan-roche:
image
Got the local grid publisher implemented as a node within the `spot_driver` package!!! Visualizing it in Foxglove in the picture above.

The publishing of local grids is configured like the pointclouds- you need to specify `publish_local_grids:=True` when you launch the driver, and since there are multiple local grids to choose from, there's an additional parameter `local_grid_name` to specify which one you want (it'll default to `obstacle_distance`)

The topic that the grids are published is constructed like the other topics in the driver- it'll prepend the name of the robot if one is specified, publishing at `/{robot name}/{grid name}`

All that's left (aside from stuff like the linting checks) is to add a check that the specified local grid name is the name of an actual grid that Spot produces, and to remove the vestigial "cutoff" logic from when the node was a part of the project I'm working on at my university robotics lab.

Might have to translate the values differently depending on what grid it is so long as it makes sense in ROS conventions- right now it just assigns one of 3 values based on some value ranges (I'm using it to estimate the footprint of objects in my project), but that doesn't make sense for a general-purpose publisher, so I'm going to have it just translate the values to whatever makes sense with the conventions of the ROS OccupancyGrid messages.

One extra thing to note is that I had to add a line to the setup script to install the `ros2_numpy` Python package I use to convert between NumPy objects and ROS messages since it doesn't have a `rosdep` binding, so it doesn't get installed by `rosdep`. It's been a godsend in my work so I'm kinda surprised it's not on `rosdep`, but I digress. The other Python packages without `rosdep` bindings are installed from a `requirements.txt` file in the `spot_wrapper` repository, but since this package is only used by the local grid publisher it didn't make sense to add it there.

---

### tl;dr
- The node has been integrated into the driver instead of being a separate package
- Launchfiles have been updated so that you just do `publish_local_grids:=True` when launching the driver, and `local_grid_name:=<grid_name>` to specify which grid to publish
- I had to add a line to the setup script to install a package directly since it didn't make sense to add it to the `requirements.txt` in the `spot_wrapper` repo since nothing in there actually uses this package

### What's left for me to do

- [ ] Remove vestigial grid probability cutoffs
- [ ] Verify the name of the requested local grid and raise an error if the grid doesn't exist
- [ ] Appease the linting gods
- [ ] Rebase commits with signoff messages (didn't realize that was required, my bad!)

ryan-roche:
I've fixed the linting and un-hardcoded the local grid that the publisher node queries from the robot, but I've run into an issue.

ROS OccupancyGrid message data can only be in `int8` format, and from the basic script I wrote to print out the data types of each of the response protobufs I get from the SDK client, they're all either `uint8` or `int16` values. I wrote two separate conversion cases, but it seems like the grid somehow fell through and was received in a different format to what I did earlier?

Regardless, I'm going to need some guidance on how to handle converting the data from the different grids into the `int8` format needed by the OccupancyGrid messages. I know that the `visualizer` example in the SDK repository reads the grid data from the protobuf in as if it's floating point values, so I think I don't fully understand the workings of the local grid data.

As of now, I'm not sure how to proceed, as the format of the data in each grid is kind of its own thing, and I'm unable to figure out an appropriate conversion from the information available in the SDK documentation.

tcappellari-bdai (approved):
Looks good to me! Thanks for making the requested edits! I just want to test this out on one of our robots before merging as a sanity check

ryan-roche:
I wouldn't merge the changes as-is- the node currently converts the grid data from the SDK response into 8-bit integers since that's the only supported format for the OccupancyGrid message data. In order for the data to still be usable, there's going to need to be some unique conversion logic for each grid- otherwise the data's just going to be clipped since most of it is in a 16-bit format

spot_driver/spot_driver/spot_local_grid_publisher.py
```diff
@@ -0,0 +1,228 @@
...
+        # "terrain_valid" and "intensity" grid protos are uint8
+
+        if raw_cells.dtype == np.uint8:
+            converted_cells = (raw_cells.astype(np.int16) - 128).astype(
+                np.int8
+            )  # Subtract a bias value so values remain correctly relative to each other
+
+        # "terrain", "no_step", and "obstacle_distance" grid protos are int16
+        else:
+            converted_cells = raw_cells.astype(np.int8)
```
khughes-bdai, line 118: Just to be clear, what lines here are the ones that are clipping data (from your earlier comment)?

spot_driver/spot_driver/spot_local_grid_publisher.py
```diff
@@ -0,0 +1,228 @@
...
+        self.get_logger().debug("📡 OccupancyGrid publisher created successfully!")
+
+        # Indicate successful initialization
+        self.get_logger().debug("[✓] Spot Local Grid Publisher Node initialized")
+
+        # Set runtime variables
+        self.first_draw_done = False
+        self.im = None
+        self.fig = None
+        self.ax = None
```
khughes-bdai, line 82: another nit -- seems like these variables aren't used, and can be deleted

PR #673, opened 2025-05-30
install renamed scripts as symlinks if --symlink-install flag is set

### LINKED ISSUE #672: Building with `--symlink-install` does not symlink renamed scripts
When building the `spot_driver` package, the `spot_ros2` and `spot_alerts` scripts are not symlinked, because the `RENAME` option in `install` forces it to be copied. This means that making changes to these scripts requires running the build again to update the code.

```cmake
install(
  PROGRAMS
  spot_driver/spot_ros2.py
  DESTINATION lib/${PROJECT_NAME}
  RENAME spot_ros2
  )
```

I think it should be possible to make this a symlink instead, but likely needs to check whether the `--symlink-install` flag is specified for colcon. I don't know if it creates a variable that's accessible from CMakeLists.

heuristicus:
## Change Overview

Fixes #672, which could be confusing if users build `spot_driver` with the `--symlink-install` flag of `colcon build`.

## Testing Done

Run `colcon build --symlink-install --packages-select spot_driver` and verify that `spot_ros` and `spot_alerts` are now symlinks. Behaviour without symlink flag should be the same.


heuristicus:
Seems that just changing to symlink may not be sufficient as launch complains about not being able to find an executable `spot_ros2`.

mhidalgo-bdai (commented):
@heuristicus would it be simpler to just rename the scripts? Executable permissions and a shebang line should do.

tcappellari-bdai (approved):
im ok with this change. Please rebase and I'll get it merged in

PR #678, opened 2025-06-09
Service for setting a pose the robot will hold

heuristicus:
## Change Overview

Adds the `/spot/posed_stand` service which allows setting the body height and roll/pitch/yaw of the robot.

## Testing Done

Tested by sending poses close to joint limits of the robot. Sending a value above the joint limits is clamped to those limits.


mhidalgo-bdai (commented):
@heuristicus how is this different from sending a stand robot command through the action interface? We have plenty of duplication across interfaces, would be good to motivate a service-based alternative.

khughes-bdai (approved):
I'm good with this change, since it seems like this is a simple enough command!

tcappellari-bdai (approved):

tcappellari-bdai:
ya im good with this. I'll merge it in once its rebased on main

PR #684, opened 2025-06-19
[SW-2907] Bazelize spot_wrapper and add python dep management

nmercaldo-bdai:
## Change Overview

Bazelized spot_wrapper as part of spot_ros2 Bazelization effort.

## Testing Done

spot_wrapper unit tests pass. 

Run unit tests with `bazel test @spot_wrapper//:all`.


nmercaldo-bdai:
@astout-bdai / any reviewers who want to join in: 

Are things outside spot_ros2 expected to depend on spot_wrapper? If so I need to ensure the spot_wrapper target is accessible via the spot_ros2 repo eg `@spot_ros2//:spot_wrapper`.

I think it is safe to assume not, put good to be sure.

PR #691, opened 2025-07-03
Allow spawning of multiple robot controllers via launch arg

gaddison-bdai:
## Change Overview

We can now specify multiple ros2_control controllers to spawn at launch by listing them, comma-separated, in the `robot_controllers` arg. Previously this arg only accepted a single controller name, allowing only a single user-specified controller from the `*_controller_config.yaml` file to be launched.

## Testing Done

This was tested by spawning two different custom controllers with different parameters using the `robot_controllers` launch arg, tested on both Mock and real Spot.

### DIFF
--- a/spot_ros2_control/launch/spot_ros2_control.launch.py
+++ b/spot_ros2_control/launch/spot_ros2_control.launch.py
@@ -163,7 +163,7 @@ def launch_setup(context: LaunchContext, ld: LaunchDescription) -> None:
                         "imu_sensor_broadcaster",
                         "foot_state_broadcaster",
                         "spot_pose_broadcaster",
-                        *robot_controllers.split(","),
+                        *robot_controllers.split(" "),
                     ],
                     namespace=spot_name,
                 )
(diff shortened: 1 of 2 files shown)


spot_ros2_control/launch/spot_ros2_control.launch.py
```diff
@@ -261,12 +262,12 @@ def generate_launch_description():
...
             ),
             DeclareLaunchArgument(
-                "robot_controller",
+                "robot_controllers",
                 default_value="forward_position_controller",
                 description=(
-                    "Robot controller to start. Must match an entry in controllers_config. For the default"
-                    " configuration file, options are forward_position_controller, forward_state_controller, or"
-                    " spot_joint_controller."
+                    "Comma-separated list of robot controller(s) to start. Each controller match an entry in"
```
khughes-bdai, line 268: super nit -- how do you feel about these being space separated instead of comma separated? I think that's the convention we use for Franka.

khughes-bdai (approved):
LGTM -- but we might need to be careful bringing this into our private repos, just to make sure that this launch arg change is carried over

spot_ros2_control/launch/spot_ros2_control.launch.py
gaddison-bdai: Yes, that would look better as well. Done [here](https://github.com/bdaiinstitute/spot_ros2/pull/691/commits/90ff1112c0903d0ebf6a0cccf8c2562d0c8b64b8).

khughes-bdai:
@gaddison-bdai I'm good with how it is. I think it makes more sense to be plural, and it should be a minor change!

PR #700, opened 2025-07-14
clang build error: removed constexpr from non-literal value (pre-C++ 20)

gaddison-bdai:
## Change Overview

When building latest main (3aa69ec) with clang-14, `spot_driver` fails to build due to `validateFrameParameter()` being marked `constexpr` but returning a non-literal value (std::optional<std::string> pre-C++20).

By removing `constexpr` from the return type of the function resolves this issue.

Here is the full build error prior to fix: 
`/home/gaddison/sporc_ws3/src/spot_ros2/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp:158:45: error: constexpr function's return type 'std::optional<std::string>' (aka 'optional<basic_string<char>>') is not a literal type
static constexpr std::optional<std::string> validateFrameParameter(const std::string& frame,
                                            ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/optional:663:7: note: 'optional<std::basic_string<char>>' is not literal because it has base class '_Optional_base<std::basic_string<char>>' of non-literal type
    : private _Optional_base<_Tp>,
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/gaddison/sporc_ws3/src/spot_ros2/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp:230:7: error: no matching function for call to 'validateFrameParameter'
      validateFrameParameter(preferred_odom_frame, kValidOdomFrameNames);
      ^~~~~~~~~~~~~~~~~~~~~~
/home/gaddison/sporc_ws3/src/spot_ros2/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp:158:45: note: candidate template ignored: substitution failure [with OptionsT = std::array<const char *const, 2>]
static constexpr std::optional<std::string> validateFrameParameter(const std::string& frame,
                                            ^
/home/gaddison/sporc_ws3/src/spot_ros2/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp:239:52: error: no matching function for call to 'validateFrameParameter'
  const std::optional<std::string> valid_tf_root = validateFrameParameter(tf_root, kValidTFRootFrameNames);
                                                   ^~~~~~~~~~~~~~~~~~~~~~
/home/gaddison/sporc_ws3/src/spot_ros2/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp:158:45: note: candidate template ignored: substitution failure [with OptionsT = std::array<const char *const, 3>]
static constexpr std::optional<std::string> validateFrameParameter(const std::string& frame,
                                            ^
3 errors generated.
`

## Testing Done

Build with colcon on ubuntu 22.04 with clang-14.

### DIFF
--- a/spot_driver/CMakeLists.txt
+++ b/spot_driver/CMakeLists.txt
@@ -1,8 +1,5 @@
 cmake_minimum_required(VERSION 3.22)
 
-set(CMAKE_CXX_STANDARD 20)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 project(spot_driver)
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/spot_driver/CMakeLists.txt
+++ b/spot_driver/CMakeLists.txt
@@ -1,5 +1,8 @@
 cmake_minimum_required(VERSION 3.22)
 
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 project(spot_driver)
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
+++ b/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
@@ -155,7 +155,8 @@ std::string getEnvironmentVariableParameterFallback(const std::shared_ptr<rclcpp
  * parameter is not a valid option, std::nullopt is returned instead.
  */
 template <typename OptionsT>
-static constexpr std::optional<std::string> validateFrameParameter(const std::string& frame, const OptionsT& base_names) {
+static constexpr std::optional<std::string> validateFrameParameter(const std::string& frame,
+                                                                   const OptionsT& base_names) {
   static_assert(type_traits::is_iterable<OptionsT>,
                 "Trait bound not satisfied for argument 'base_names', type not iterable.");
   static_assert(std::is_convertible_v<typename OptionsT::value_type, std::string>,
--- a/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
+++ b/spot_driver/src/interfaces/rclcpp_parameter_interface.cpp
@@ -155,7 +155,7 @@ std::string getEnvironmentVariableParameterFallback(const std::shared_ptr<rclcpp
  * parameter is not a valid option, std::nullopt is returned instead.
  */
 template <typename OptionsT>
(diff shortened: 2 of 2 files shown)


gaddison-bdai:
Setting C++ version to 20 does, in fact, fix the issue, even with the original constexpr [here](https://github.com/bdaiinstitute/spot_ros2/pull/700/commits/56e6e4b6c01e25fed06bedb2fa54c1cdf7d34c1c).

@mbrauckmann-bdai @jcarpinelli-bdai do y'all forsee any issues with packages depending on `spot_driver` that may have used a different C++ version? I assume not, since other packages in this repo (i.e. `spot_ros2_control`) already specify C++20.

mhidalgo-bdai:
> IMO compiling with c++20 should be safe here but I'd make sure spot_ros2 codeowners are okay with it.

All ROS 2 distributions are C++17 compatible, see [REP-2000](https://www.ros.org/reps/rep-2000.html), so I think it would've been fine to use C++20 if we were not exporting targets downstream. We are. 

Based on where we use `validateFrameParameter`, I don't think dropping `constexpr` will have measurable impact in performance. CC @khughes-bdai @tcappellari-bdai for input.

khughes-bdai:
Just catching up on this! @gaddison-bdai I would lean towards dropping the `contexpr` from `validateFrameParameter` since that seems like an easy fix, then we won't need to modify anything in the cmakelists. 
I realize we are doing the same thing in the spot hardware interface cmakelists [here](https://github.com/bdaiinstitute/spot_ros2/blob/3aa69ec608b7f60909b4fbbae46867043da50b25/spot_hardware_interface/CMakeLists.txt#L6). This was so we could use `jthread` for the state streaming interface, and to my knowledge we haven't had any issues because of this internally or with external users. But it seems preferable to try to avoid this in this package if possible just to keep things simple :)

gaddison-bdai:
Indeed, removing `constexpr` does solve the clang build issue, and seems to have less potential to cause issues with other packages.

> Just catching up on this! @gaddison-bdai I would lean towards dropping the `contexpr` from `validateFrameParameter` since that seems like an easy fix, then we won't need to modify anything in the cmakelists. I realize we are doing the same thing in the spot hardware interface cmakelists [here](https://github.com/bdaiinstitute/spot_ros2/blob/3aa69ec608b7f60909b4fbbae46867043da50b25/spot_hardware_interface/CMakeLists.txt#L6). This was so we could use `jthread` for the state streaming interface, and to my knowledge we haven't had any issues because of this internally or with external users. But it seems preferable to try to avoid this in this package if possible just to keep things simple :)

That makes sense. I removed the `constexpr` and removed the requirement for C++20. Also just tested and build works with clang now! Let me know how it looks now @mhidalgo-bdai @khughes-bdai

khughes-bdai (approved):
LGTM!

PR #712, opened 2025-07-28
Make command polling rates configurable

mhidalgo-bdai:
## Change Overview

`spot_driver` polls quite a bit. BD's Spot SDK doesn't allow any better. However, some polling rates are unnecessarily high, which can take a toll on the receiving end (the actual robot, a simulation of it, etc.). This patch makes these rates configurable and normalizes them to a single, reasonable default. 

## Testing Done

Tested in simulation. A double check on hardware would be good too.


khughes-bdai (approved):
LGTM, ran a few examples on robot and everything seems normal

Issue #29, opened 2023-03-31
Odometry velocities reported in wrong frame


Jannkar:
According to [msg/Odometry documentation](https://docs.ros2.org/foxy/api/nav_msgs/msg/Odometry.html), linear and angular velocities in Odometry message should be reported relative to child_frame_id.

Currently the velocities are reported in `odom` frame, which is causing issues when using this data with other existing ROS packages.

This is how the velocities are currently set in `ros_helpers.py`:
```
def GetOdomTwistFromState(state, spot_wrapper):
    """Maps odometry data from robot state proto to ROS TwistWithCovarianceStamped message
    Args:
        data: Robot State proto
        spot_wrapper: A SpotWrapper object
    Returns:
        TwistWithCovarianceStamped message
    """
    twist_odom_msg = TwistWithCovarianceStamped()
    local_time = spot_wrapper.robotToLocalTime(state.kinematic_state.acquisition_timestamp)
    twist_odom_msg.header.stamp = Time(sec=local_time.seconds, nanosec=local_time.nanos)
    twist_odom_msg.twist.twist.linear.x = state.kinematic_state.velocity_of_body_in_odom.linear.x
    twist_odom_msg.twist.twist.linear.y = state.kinematic_state.velocity_of_body_in_odom.linear.y
    twist_odom_msg.twist.twist.linear.z = state.kinematic_state.velocity_of_body_in_odom.linear.z
    twist_odom_msg.twist.twist.angular.x = state.kinematic_state.velocity_of_body_in_odom.angular.x
    twist_odom_msg.twist.twist.angular.y = state.kinematic_state.velocity_of_body_in_odom.angular.y
    twist_odom_msg.twist.twist.angular.z = state.kinematic_state.velocity_of_body_in_odom.angular.z
    return twist_odom_msg
```

heuristicus:
This has also been reported at https://github.com/heuristicus/spot_ros/issues/95

kjvanteeffelen:
Hello, 

Are there any plans to add the fix to the current code? Indeed, the Twist definition in odometry is 'wrong' according to the ROS standards. This will lead to problems when for instance using the robot_localization package with Spot. 

Until it's fixed, I'm using an additional node which subscribes to the current odometry, applies the fix and publishes it to the '/odometry_corrected' topic. This could easily be integrated into the code directly I believe. It was largely copied from https://github.com/heuristicus/spot_ros/issues/95. 

```python
import rclpy
import copy
[...]

PR #416, opened 2024-06-25
[SW-893] odometry velocities reported in the wrong frame

khughes-bdai:
## Change Overview

As brought up in https://github.com/bdaiinstitute/spot_ros2/issues/29, the odometry twist should be reported in the body frame, not the odom frame. This will make the spot driver easier to use with other ROS tools that expect this data to be in this format. 

## Testing Done

- [x] walk the robot in a straight line from a variety of different heading angles. ensure that the twist component has linear velocity only in the x dimension.
- [x] update tests to account for new change
- [ ] I want confirmation from users in https://github.com/bdaiinstitute/spot_ros2/issues/29 that this correctly resolves the issue before merging (leaving the do not merge label on until this is verified)

PR #738, opened 2025-09-09
Run corrected_odom script to provide odometry velocity on robot frame

alisys-robotics:
## Change Overview

Temporal fix for #29: here I'm just pushing the script created by @kjvanteeffelen on 2023 and running it as part of the Spot driver.
It works fine, but for sure the proper solution is to patch the driver itself.

## Testing Done

We run this script for several weeks as part of our Spot setup


khughes-bdai:
Hey @alisys-robotics, I have a PR up that modifies the odom topic at the source: https://github.com/bdaiinstitute/spot_ros2/pull/416
Are you able to test if this resolves the issue?

PR #780, opened 2025-12-30
Fix headers: cv_bridge.hpp migration & gmock include cleanup

maeri18:
## Change Overview

- changed cv_bridge.h imports to cv_bridge.hpp imports  in these files : 

1. spot_driver/src/api/default_image_client.cpp
2. spot_driver/src/conversions/decompress_images.cpp
3. spot_driver/src/image_stitcher/image_stitcher.cpp

- Commented gmock-generated-matchers.h import in spot_driver/test/include/spot_driver/matchers.hpp, as gmock-generated-matchers is not part of new versions of gmock.

### DIFF
--- a/spot_driver/src/api/default_image_client.cpp
+++ b/spot_driver/src/api/default_image_client.cpp
@@ -4,9 +4,9 @@
 
 #include <bosdyn/api/directory.pb.h>
 #include <bosdyn/api/image.pb.h>
-#include <cv_bridge/cv_bridge.h>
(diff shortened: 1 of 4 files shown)


mhidalgo-rai:
@maeri18 perhaps we can use [`__has_include` ](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html). It works for both `gcc` and `clang` nowadays, and it would spare us the per distribution branching strategy (for now at least).

Issue #783, opened 2026-02-02
Arm checks are problematic


mhidalgo-rai:
Spot ROS 2 exposes arm availability information over a service interface that downstream code must call. This has proven problematic over time, as early service calls tend to be somewhat unreliable on Humble middlewares: timeouts, races. Long timeouts and retry policies only go so far. Spot ROS 2 would be better off with a transient local topic distributing this information.

PR #785, opened 2026-02-03
Publish transient local arm availability flag

mhidalgo-rai:
## Change Overview

Closes https://github.com/bdaiinstitute/spot_ros2/issues/783. Precisely what the title says. As an alternative to node parameter checks, which are known to be flaky (but kept for backwards compatibility).

## Testing Done

Tested on a mock driver using `ros2 topic echo`.


khughes-bdai (approved):
LGTM!

PR #670, opened 2025-05-23
Improve namespacing handling in spot ros2 control

khughes-bdai:
## Change Overview

Please include a summary of the changes made and the relevant ticket or issue.

## Testing Done

Please create a checklist of tests you plan to do and check off the ones that have been completed successfully. Ensure that ROS 2 tests use `domain_coordinator` to prevent port conflicts. Further guidance for testing can be found on the [ros utilities wiki](https://github.com/bdaiinstitute/ros_utilities/wiki/Testing-guidelines).

PR #794, opened 2026-04-17
Fix spot_imu_broadcaster_parameters missing from link and install

jakl-bdai:
`spot_imu_broadcaster_parameters` was added via `generate_parameter_library` in #670 but was never added to `target_link_libraries` or `install(TARGETS ...)`. This causes the generated header to never be built, breaking fresh builds with:

```
fatal error: spot_controllers/spot_imu_broadcaster_parameters.hpp: No such file or directory
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,7 @@ jobs:
       - name: Build packages
         run: |
           source /opt/ros/$ROS_DISTRO/setup.bash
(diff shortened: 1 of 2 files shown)


jbarry-bdai (approved):

mhidalgo-rai (approved):
Great catch @jakl-bdai

taughz (approved):

PR #808, opened 2026-06-29
Update get_name_and_prefix with upstream PathJoinSubstitution change

khughes-bdai:
## Change Overview

`PathJoinSubstitution` recently had a sync that broke how the spot name and TF prefix were assembled, causing the TF prefix to not have a "/" in it when `spot_name` was set. ([before](https://github.com/ros2/launch/blob/e716d1fcc719b4155b7c99f2ae25cc3466634231/launch/launch/substitutions/path_join_substitution.py) vs [after](https://github.com/ros2/launch/blob/cf96072c2045e821eecf43c80c326ab74cb74b7a/launch/launch/substitutions/path_join_substitution.py)). 
* expected: `spot_name:=Spot` -> TF prefix is `Spot/`
* actual: `spot_name:=Spot` -> TF prefix is `Spot` 

This change uses a `PythonExpression` instead of `PathJoinSubstitution` to ensure that the tf prefix is created properly

## Testing Done
- [x] Unit tests pass
- [x] Tested with and without spot_name launch arg set in `spot_driver.launch.py`, matches expected behavior: 
  - [x] `spot_name` is unset -> TF prefix is empty string
  - [x] `spot_name:=Spot` -> TF prefix is `Spot/`

### DIFF
--- a/spot_common/spot_common/launch/spot_launch_helpers.py
+++ b/spot_common/spot_common/launch/spot_launch_helpers.py
@@ -8,7 +8,7 @@ from typing import Any, Dict, List, Literal, Optional, Tuple, Union
 import yaml
 from launch import LaunchContext, Substitution
 from launch.actions import DeclareLaunchArgument
-from launch.substitutions import PathJoinSubstitution
+from launch.substitutions import PythonExpression
 
 from spot_wrapper.wrapper import SpotWrapper
 from synchros2.launch.actions import DeclareBooleanLaunchArgument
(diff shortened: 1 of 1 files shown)


tcappellari-bdai (approved):

spot_common/spot_common/launch/spot_launch_helpers.py
```diff
@@ -312,7 +312,7 @@ def get_name_and_prefix(ros_params: Dict[str, Any]) -> Tuple[Union[str, Substitu
     tf_prefix: Optional[Union[str, Substitution]] = ros_params[_FRAME_PREFIX] if _FRAME_PREFIX in ros_params else None
     if tf_prefix is None:
         if isinstance(spot_name, Substitution):
-            tf_prefix = PathJoinSubstitution([spot_name, ""])
+            tf_prefix = PythonExpression(['"', spot_name, '" + "/"'])
```
mhidalgo-rai, line 315: @khughes-bdai nit^N!: something like:
```suggestion
            tf_prefix = TextSubstitution([spot_name, "/"])
```
should work just the same.

mhidalgo-rai (approved):

spot_common/spot_common/launch/spot_launch_helpers.py
khughes-bdai: This wasn't working on my end as TextSubstitution expects only string literals in it's constructor :(

mhidalgo-rai: Ahh, it should've been `[spot_name, "/"]`, lists of `Substitution`s are `Substitution` IIRC.

Anyways, this works too.

PR #812, opened 2026-07-30
fix(ci): protect durable image tags from GHCR retention pruning

brennan-vanderlaan-bd-ai:
### Broke
`docker pull ...:main` → `manifest unknown`. The weekly retention job (`snok/container-retention-policy@v2`, `skip-tags: main`) only protected the tagged **index** manifest, not the untagged **per-arch child manifests** of the multi-arch image. Once those children aged past `cut-off`, the pruner deleted them → `main` index points at missing manifests.

### Fix
- Upgrade to `container-retention-policy@v3.1.0` (auto-protects multi-arch children of kept tags, fails closed).
- Before pruning, resolve every digest reachable from each **durable** tag (index + arch children + attestations) and pass via `skip-shas`. Durable = all tags except ephemeral `pr-*`. Protects `main` + release tags even during quiet periods > cut-off.
- Works with default `GITHUB_TOKEN` (no PAT needed — we enumerate/filter tags ourselves since the action's tag operators require a PAT).
- Adds `workflow_dispatch` (defaults to `dry-run`).

### Verify
Run manually with dry-run first; confirm only `pr-*` are selected before the Sunday schedule.

### DIFF
--- a/.github/workflows/maintenance.yml
+++ b/.github/workflows/maintenance.yml
@@ -3,24 +3,126 @@ name: spot_ros2 CI maintenance
 on:
   schedule:
     - cron: "0 0 * * 0"  # once a week
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: "Run the prune in dry-run mode (log deletions without performing them)"
+        type: boolean
+        default: true
 
 env:
   ORG_NAME: rai-opensource
   # github.repository as <account>/<repo>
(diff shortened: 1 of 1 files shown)


taughz (approved):

taughz (changes requested):
Testing, please ignore...

taughz (approved):

